PR #115, opened 2025-08-27
Refactor `positronic.dataset.Signal` core; unify indexing and time semantics

vertix:
Summary

- Centralizes all indexing/time semantics in core Signal.
- Updates vector/video to implement minimal abstract interface.
- Consolidates tests: core covers semantics; signal/video focus on I/O.
- Clarifies and documents behavior in code and README.

Key changes

- New interface in implementations:
    - len, _ts_at, _values_at, _search_ts (floor index; -1 if before-first).
- Base Signal now handles:
    - Indexing: ints, positive-step slices, integer arrays (no boolean masks).
    - Time scalar: at-or-before; KeyError if < first.
    - Time window (non-stepped): [start, stop); empty signal → empty; start < first intersects to [first, stop); inject carried-back at start when
between records.
    - Time window (stepped): [start, stop:step]; step>0; start required; start < first → KeyError; timestamps are requested ones.
    - Time arrays: empty → empty view; integer dtype required; any ts < first → KeyError.

Implementations

- vector.py: implements the abstract methods; _search_ts returns empty array for empty input; TypeError on non-integer dtype.
- video.py: implements the abstract methods; decodes frames in _values_at; same empty/_search_ts behavior.

Tests

- Expanded core tests for:
    - array-on-slice, nested time-slices, stepped iteration, mixed-before-first (KeyError), empty time arrays (empty view).
- Pruned signal/video tests to content and writer I/O only; added minimal interface tests for vector/video to exercise len, _ts_at, _values_at,
_search_ts.


positronic/dataset/core.py
```diff
@@ -91,7 +155,140 @@ def __getitem__(self, index_or_slice: int | slice | Sequence[int] | np.ndarray)
             If index: Tuple of (value, timestamp_ns)
             If slice/array-like: Signal[T] view of the original data
         """
-        pass
+        match index_or_slice:
+            case int() as idx:
```
JIy3AHKO, line 159: This branch will not capture `np.intX` types. This could happen if your index came from np.ndarray:

```python
import numpy as np
idx = np.arange(10)
match idx[0]:
    case int() as idx:
        # this branch will not be reached since idx[0] is np.int64
        print(idx)
```

positronic/dataset/README.md
```diff
@@ -21,35 +21,30 @@ __Dataset__ – ordered collection of Episodes with sequence-style access (index
...
-    #     precedes the first record, a KeyError is raised. step must be positive.
-    # * signal.time[[t1, t2, ...]] returns a `Signal` sampled at the provided timestamps.
-    #     Each element is (value_at_or_before_t_i, t_i). Raises KeyError if any t_i precedes
-    #     the first record.
-    @property
-    def time(self):
-        pass
+    # Minimal abstract interface (implementations must provide):
+    def __len__(self) -> int: ...                # number of records
+    def _ts_at(self, idx_or_indices) -> int | np.ndarray: ...
```
JIy3AHKO, line 29: Maybe we could simplify this, but changing `idx_or_indices` to just `indices`?

We could implement the following logic once:
```pseudo
if idx is singular index:
   idx = [idx]
   result = base_signal[idx]
   return result[0]
```

This way derived classes only need to worry about implementing single idx case - `Sequence[int]`.

positronic/dataset/core.py
```diff
@@ -91,7 +155,140 @@ def __getitem__(self, index_or_slice: int | slice | Sequence[int] | np.ndarray)
             If index: Tuple of (value, timestamp_ns)
             If slice/array-like: Signal[T] view of the original data
         """
-        pass
+        match index_or_slice:
```
JIy3AHKO, line 158: It feels, like this logic belongs to `_SignalView` class.

1. It returns `_SignalView`
2. I think there should be no difference from outside between `original` and `sliced` signal.

If we would make an analogy with `numpy`:
1. `Signal` interface looks more like a base data (like arr.data - in numpy case just a pointer to a memory)
2. `_SignalView` is high-level wrapper which provides different access modes to this data (like np.ndarray itself)

So I think we should make current `Signal` class a "hidden implementation detail" and `_SignalView` a public class

positronic/dataset/video.py
```diff
@@ -451,48 +250,31 @@ def _get_frame_at_index(self, index: int) -> Tuple[np.ndarray, int]:
...
+        else:
+            idxs = np.asarray(index_or_indices, dtype=np.int64)
+        return [self._get_frame_at_index(int(i))[0] for i in idxs]
 
-    def _p_idx_for_range(self, start_idx: int, stop_idx: int) -> Tuple["VideoSignal", np.ndarray]:
-        """Map absolute range [start_idx:stop_idx] to parent indices for slice-based time lookups."""
-        return self, np.arange(start_idx, stop_idx, dtype=np.int64)
+    def _search_ts(self, ts_or_array: int | Sequence[int] | np.ndarray):
+        self._load_timestamps()
+        if isinstance(ts_or_array, (int, np.integer)):
```
JIy3AHKO, line 273: timestamps could be float since we just finding the closes ts < target_ts

JIy3AHKO (changes requested):

PR #119, opened 2025-09-02
IK management mode for Franka arm

vertix:


positronic/geom/__init__.py
```diff
@@ -377,6 +377,8 @@ def as_rotvec(self) -> np.ndarray:
                           of the vector indicates the axis of rotation and its magnitude
                           represents the angle in radians.
         """
+        if self[0] > 1.0 or self[0] < -1.0:
```
JIy3AHKO, line 380: Shouldn't we call `geom.normalise_quat` in Rotation `__init__`  to prevent this situation? In normalized quat there shouldn't be any values |v| >= 1

Your current fix could possibly break in some other scenarios. (e. g. in `angle` property)

positronic/drivers/roboarm/franka.py
```diff
@@ -51,13 +53,41 @@ def encode(self, q, dq, ee_pose):
...
+
+
+def _damped_pinv(J: np.ndarray, lambda2: float) -> np.ndarray:
+    # J: 6x7, return 7x6 damped pseudo-inverse using J^T (J J^T + λ I)^-1
+    I6 = np.eye(6)
+    JJt = J @ J.T
+    return J.T @ np.linalg.inv(JJt + lambda2 * I6)
+
+
+class CartesianMode(Enum):
```
JIy3AHKO, line 78: Maybe we should name it `IKSolver`? 

```python
class IKSolver(Enum):
   FRANKY = 0
   POSITRONIC = 1
```

PR #124, opened 2025-09-09
Add FastAPI web server for Positronic dataset visualization

vertix:
Already useful for checking datasets:

`python -m positronic.server.positronic_server --root=/Users/vertix/Documents/datasets/stack-cubes-pimm`

### DIFF
--- a/positronic/server/positronic_server.py
+++ b/positronic/server/positronic_server.py
@@ -1,0 +1,207 @@
+"""A FastAPI web server for visualizing Positronic LocalDatasets using Rerun."""
+
+from __future__ import annotations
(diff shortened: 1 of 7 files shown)


positronic/server/positronic_server.py
```diff
@@ -0,0 +1,207 @@
+"""A FastAPI web server for visualizing Positronic LocalDatasets using Rerun."""
+
+from __future__ import annotations
```
JIy3AHKO, line 3: Same here

JIy3AHKO (changes requested):

PR #152, opened 2025-09-26
Add a movement sensitivity parameter

alexaibio:
Add a new parameter sensitivity_scale to cfg/webxr.py which changes sensitivity of movement during simulation with mobile device and webxr

### DIFF
--- a/positronic/drivers/webxr.py
+++ b/positronic/drivers/webxr.py
@@ -123,7 +123,7 @@ class WebXR(pimm.ControlSystem):
                  ssl_certfile: str = "cert.pem",
                  frontend: str = "oculus",
                  use_https: bool = True,
-                 sensitivity_scale: float = 1.0):
+                 sensitivity: float = 1.0):
         self.port = port
         self.ssl_keyfile = ssl_keyfile
         self.ssl_certfile = ssl_certfile
@@ -134,7 +134,7 @@ class WebXR(pimm.ControlSystem):
         self.frame = pimm.ControlSystemReceiver(self)
         self.controller_positions = pimm.ControlSystemEmitter(self)
         self.buttons = pimm.ControlSystemEmitter(self)
-        self.sensitivity_scale = sensitivity_scale
+        self.sensitivity = sensitivity
 
     def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[pimm.Sleep]:  # noqa: C901
         app = FastAPI()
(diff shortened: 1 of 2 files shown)


positronic/drivers/webxr.py
```diff
@@ -42,13 +42,13 @@
...
-def _parse_controller_data(data: dict):
+def _parse_controller_data(data: dict, scale: float = 1.0):
     controller_positions = {'left': None, 'right': None}
     buttons_dict = {'left': None, 'right': None}
     for side in ['right', 'left']:
         if data['controllers'][side] is not None:
-            translation = np.array(data['controllers'][side]['position'], dtype=np.float64)
-            rotation = np.array(data['controllers'][side]['orientation'], dtype=np.float64)
+            translation = np.array(data['controllers'][side]['position'], dtype=np.float64) * scale
+            rotation = np.array(data['controllers'][side]['orientation'], dtype=np.float64) * scale
```
vertix, line 51: I think this is not what you want, due to several reasons:
* This is quaternion representation of rotations (https://www.youtube.com/watch?v=zjMuIxRvygQ), so multiplying just changes magnitude of quaternion, but not the actually rotation it represents.
* Even if doing it's right (via quaternion power), this will amplify your rotations, which might feel bad.

So long story short, I suggest to amplify only translation, as this will make robots more sensitive to movements, but not to rotations.

positronic/drivers/webxr.py
```diff
@@ -122,7 +122,8 @@ def __init__(self,
                  ssl_keyfile: str = "key.pem",
                  ssl_certfile: str = "cert.pem",
                  frontend: str = "oculus",
-                 use_https: bool = True):
+                 use_https: bool = True,
+                 sensitivity_scale: float = 1.0):
```
vertix, line 126: Let's call it just `sensitivity`, here and across configs.

positronic/drivers/webxr.py
```diff
@@ -204,7 +206,7 @@ async def websocket_endpoint(websocket: WebSocket):
                     try:
                         # Use asyncio.wait_for with timeout to avoid blocking indefinitely
                         data = await asyncio.wait_for(websocket.receive_json(), timeout=1)
-                        controller_positions, buttons = _parse_controller_data(data)
+                        controller_positions, buttons = _parse_controller_data(data, self.sensitivity_scale)
```
vertix, line 209: `_parse_controller_data` has nothing to do with the scale, it is about parsing. Scaling should be done here.

You can do
```
controller_positions, buttons = _parse_controller_data(data)
controller_positions.translation *= self.sensitivity_scale
```

vertix (changes requested):

PR #345, opened 2026-03-13
Add SPA episode navigation to avoid rerun WASM reload

vertix:
## Summary
- Replace full page reloads with in-place RRD data swapping via rerun's `WebHandle` API (`remove_receiver`/`add_receiver`), keeping the 35MB WASM alive across episode navigation
- Add a Service Worker with cache-first strategy for `/static/rerun/` assets, so the WASM persists across hard refreshes even with self-signed certs
- Add `/api/episode/{id}` JSON endpoint for SPA metadata fetching
- Fix sidebar content accumulating on repeated navigation
- Add `jinja2` dependency (required by starlette's `Jinja2Templates`)

## Test plan
- Navigate between episodes using prev/next buttons and verify rerun viewer updates without full reload
- Hard refresh the page and verify WASM loads from Service Worker cache
- Use browser back/forward and verify correct episode loads
- Test filtered episode navigation from the episodes table
- Dismiss filter and verify nav controls restore correctly

### DIFF
--- a/positronic/server/positronic_server.py
+++ b/positronic/server/positronic_server.py
@@ -103,6 +103,17 @@ async def cache_rerun_assets(request: Request, call_next):
     return response
 
 
+def _make_serializable(obj):
+    """Ensure obj is JSON serializable (e.g. convert datetime)."""
+    if isinstance(obj, datetime):
+        return obj.isoformat()
+    if isinstance(obj, dict):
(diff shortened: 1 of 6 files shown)


vertix:
Closing unmerged — too far behind to rebase. `positronic_server.py` and `episode.html` have been reworked repeatedly since (#362, #467, #505, #539, #616). The `jinja2` dependency landed independently.

PR #374, opened 2026-03-24
Fix DreamZero full finetune h100x8 checkpoint disk overflow

vertix:
## Summary
- Make `save_total_limit` configurable (default 10, overridable per preset)
- Use `+training_args.save_only_model=true` for full finetune h100x8 presets — DeepSpeed checkpoints are ~200GB each (model + optimizer state across 8 ranks), exceeding disk with `save_total_limit=10`
- LoRA h100x8 presets unchanged — LoRA checkpoints are small enough for full DeepSpeed saves

## Test plan
- Tested wan2.2 full finetune on 8×H100: training runs past checkpoint saves without disk-full crash
- Inference tested with checkpoint-2000 from `w22f8_230326` — server loads and serves correctly

### DIFF
--- a/positronic/vendors/dreamzero/train.py
+++ b/positronic/vendors/dreamzero/train.py
@@ -61,6 +61,7 @@ def _dreamzero_root():
     dataloader_num_workers=1,
     deepspeed=None,
     resume=False,
+    save_total_limit=10,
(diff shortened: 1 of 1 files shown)


vertix:
@codex review

PR #389, opened 2026-05-14
Add dataset sampling for LeRobot conversion and `droid_spoons` ablation setup

vertix:
## Summary
- Add `share` (default 1.0) and `seed` (default 42) parameters to both `to_lerobot.py` conversion scripts to deterministically subsample episodes when converting Positronic datasets to LeRobot format
- Add `droid_spoons` config in `positronic/cfg/ds/internal.py` — `droid_clean` filtered to only spoons-task episodes (167/449 episodes)
- Add `docker/docker-compose.spoons-ablation.yml` for serving the resulting GR00T checkpoints
- Broaden pre-commit `check-yaml` exclude pattern to cover all `docker-compose.<suffix>.yml` files (not just `phail`)

## Test plan
- Verified `droid_spoons` produces 167 episodes (vs 449 for `droid_clean`)
- Used the new `--share` to convert 100% / 50% / 25% subsets of `droid_spoons` to GR00T's `ee_rot6d` interim format, then trained three GR00T-N1.6 jobs to 150k steps on H100 VMs. All three completed and saved checkpoints to `s3://checkpoints/droid_spoons/groot/ee_rot6d_rel_{100,50_v2,25}/`
- Sampling determinism: same `seed` produces same indices on re-runs (50% subset is nested inside the 100% set)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
-        exclude: docker-compose\.phail\..*yml  # Uses Docker Compose !override tags
+        exclude: docker-compose\..+\.yml  # Uses Docker Compose !override tags
       - id: check-toml
       - id: check-added-large-files
         args: ['--maxkb=1000']
--- a/positronic/cfg/ds/internal.py
+++ b/positronic/cfg/ds/internal.py
@@ -177,6 +177,12 @@ def droid_clean(dataset):
 
 droid_clean = droid_clean.override(dataset=droid)
 
+
+@cfn.config(dataset=droid_clean)
+def droid_spoons(dataset):
+    return FilterDataset(dataset, lambda ep: ep.static.get('task') == SPOONS_TASK)
+
+
 droid_recovery = droid_clean.override(
     dataset=droid.override(recovery_all=True, recovery_towels=True, duplicate_recovery=False)
 )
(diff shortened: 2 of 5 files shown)


positronic/cfg/ds/internal.py
```diff
@@ -177,6 +177,12 @@ def droid_clean(dataset):
 
 droid_clean = droid_clean.override(dataset=droid)
 
+
+@cfn.config(dataset=droid_clean)
```
vertix, line 183: Please add a comment that this dataset is for ablation experiment on data size

.pre-commit-config.yaml
```diff
@@ -16,7 +16,7 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
-        exclude: docker-compose\.phail\..*yml  # Uses Docker Compose !override tags
+        exclude: docker-compose\..+\.yml  # Uses Docker Compose !override tags
```
vertix, line 19: Can you explain this change?

positronic/cfg/ds/internal.py
vertix: Done — added a comment above the definition explaining it's used for the data-size ablation (combined with `--share`/`--seed` in the conversion CLI).

.pre-commit-config.yaml
vertix: The existing exclude pattern only covered `docker-compose.phail.*.yml`, but the new `docker-compose.spoons-ablation.yml` also uses Docker Compose's `!override` tag — which is not standard YAML and trips up `check-yaml` (PyYAML doesn't know how to construct it). Broadening the pattern to `docker-compose\..+\.yml` covers any non-base compose variant. The base `docker-compose.yml` (no inner suffix) still gets checked.

PR #394, opened 2026-05-18
Add support for authenticating to remote servers

RikHeijdens:
## Summary

Add support for setting custom headers on requests to the server to enable requests to secured server endpoints.

Example usage:

```
uv run positronic-inference sim \
     --policy=.remote \
     --policy.host=<workspace> \
     --policy.port=443 --policy.secure=True \
     --policy.headers='{"Key": "<VALUE>"}'
```

### DIFF
--- a/positronic/offboard/client.py
+++ b/positronic/offboard/client.py
@@ -80,8 +80,7 @@ class InferenceClient:
     def __init__(self, host: str, port: int, *, headers: dict[str, str] | None = None, secure: bool = False):
         self.host = host
         self.port = port
-        self.secure = secure
-        self.headers = dict(headers) if headers else {}
+        self.headers = dict(headers) if headers else None
         ws_scheme = 'wss' if secure else 'ws'
         http_scheme = 'https' if secure else 'http'
         default_port = 443 if secure else 80
@@ -111,6 +110,6 @@ class InferenceClient:
 
     def list_models(self) -> list[str]:
         """List available models from the server."""
-        response = httpx.get(f'{self.api_url}/models', headers=self.headers or None)
+        response = httpx.get(f'{self.api_url}/models', headers=self.headers)
         response.raise_for_status()
         return response.json()['models']
--- a/positronic/offboard/client.py
+++ b/positronic/offboard/client.py
@@ -77,11 +77,17 @@ class InferenceSession:
 
 
 class InferenceClient:
-    def __init__(self, host: str, port: int):
+    def __init__(self, host: str, port: int, *, headers: dict[str, str] | None = None, secure: bool = False):
         self.host = host
         self.port = port
-        self.base_uri = f'ws://{host}:{port}/api/v1/session'
-        self.api_url = f'http://{host}:{port}/api/v1'
+        self.secure = secure
+        self.headers = dict(headers) if headers else {}
+        ws_scheme = 'wss' if secure else 'ws'
+        http_scheme = 'https' if secure else 'http'
+        default_port = 443 if secure else 80
+        netloc = host if port == default_port else f'{host}:{port}'
+        self.base_uri = f'{ws_scheme}://{netloc}/api/v1/session'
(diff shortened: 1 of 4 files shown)


positronic/offboard/client.py
```diff
@@ -77,11 +77,17 @@ def close(self):
 
 
 class InferenceClient:
-    def __init__(self, host: str, port: int):
+    def __init__(self, host: str, port: int, *, headers: dict[str, str] | None = None, secure: bool = False):
         self.host = host
         self.port = port
-        self.base_uri = f'ws://{host}:{port}/api/v1/session'
-        self.api_url = f'http://{host}:{port}/api/v1'
+        self.secure = secure
```
vertix: Looks like `self.secure` is not used anywhere, no need to introduce it

positronic/offboard/client.py
```diff
@@ -77,11 +77,17 @@ def close(self):
...
+        self.headers = dict(headers) if headers else {}
```
vertix: I think `None` as default value is better, as then in line 114, it is just
`response = httpx.get(f'{self.api_url}/models', headers=self.headers)`

positronic/offboard/client.py
```diff
@@ -94,14 +100,17 @@ def new_session(self, model_id: str | None = None, open_timeout: float = 10.0) -
                         Model loading timeout is controlled by per-message timeout in handshake.
         """
         uri = self.base_uri if model_id is None else f'{self.base_uri}/{model_id}'
+        connect_kwargs: dict[str, object] = {'open_timeout': open_timeout}
+        if self.headers:
+            connect_kwargs['additional_headers'] = self.headers
         try:
-            ws = connect(uri, open_timeout=open_timeout)
+            ws = connect(uri, **connect_kwargs)
```
vertix: Testing against the Modal endpoint I discovered that when the container is scaled to zero, Modal holds the WS upgrade open through the cold start, the client's `open_timeout=10.0` fires, and `new_session()` dies with `TimeoutError`: timed out while waiting for handshake response. So every first connect after idle fails unless you manually warm the endpoint first.

Could we retry the open until a deadline? Roughly:

```python
def new_session(self, model_id=None, open_timeout=10.0, connect_deadline=180.0) -> InferenceSession:
    uri = self.base_uri if model_id is None else f'{self.base_uri}/{model_id}'
    connect_kwargs: dict[str, object] = {'open_timeout': open_timeout}
    if self.headers:
        connect_kwargs['additional_headers'] = self.headers

    deadline = time.monotonic() + connect_deadline
    backoff = 1.0
    while True:
        try:
            ws = connect(uri, **connect_kwargs)
            break
        except TimeoutError as e:
            if time.monotonic() >= deadline:
                raise TimeoutError(f'{e} (connecting to {self.host}:{self.port})') from e
            logger.info('Server not ready (cold start?); retrying in %.0fs', backoff)
            time.sleep(backoff)
            backoff = min(backoff * 2, 30.0)
        except OSError as e:
            raise type(e)(f'{e} (connecting to {self.host}:{self.port})') from e
    return InferenceSession(ws)
```

vertix (changes requested):
Thank you for your contribution! Before we merge, please confirm (just reply "I agree") that:
- You have the right to submit this code – it's your own work, or you have permission from whoever does.
- You agree to license this contribution and future contributions to this repository under the project's license (Apache-2.0, see LICENSE).

RikHeijdens:
> Thank you for your contribution! Before we merge, please confirm (just reply "I agree") that:
> 
> * You have the right to submit this code – it's your own work, or you have permission from whoever does.
> * You agree to license this contribution and future contributions to this repository under the project's license (Apache-2.0, see LICENSE).

I agree

vertix:
@RikHeijdens Sorry for trouble, I recently merged another PR that introduced a conflict. It should be easy to resolve though

vertix (approved):

PR #400, opened 2026-05-19
Add reproducible behavioral golden for the inference pipeline

vertix:
## Summary
- Adds `positronic/policy/tests/test_golden_pipeline.py` + `golden_pipeline.json`: a deterministic, bit-reproducible golden that locks the robot-facing behavior of the full inference chain (policy → codec → `Harness` → timestamp-respecting driver → `DsWriterAgent`).
- The golden captures the **recorded episode state** (`robot_state.ee_pose`, `robot_state.q`, `grip`) at the `DsWriterAgent` output — the representation-free *effect* of the pipeline. Value regressions change `value`; timing/anchoring regressions move `ts_ns`; horizon/gating regressions diverge the trajectory.
- One scenario exercises reactive inference cycles, chunk + `ActionHorizon` truncation, nonzero deterministic inference latency (anchoring), an injected `RobotStatus.ERROR` → `Recover` → resume, and the RUN/FINISH lifecycle.
- Runs entirely on CPU with `MockClock` only — no GL/GPU/MuJoCo and no wall clock in the asserted path.
- Renames the harness parameter `simulate_timeout` → `simulate_inference` (`bool | float`): `True` measures real wall latency, a `float` advances the sim clock by exactly that many seconds (used by the golden), `False` keeps inference free. Updates `inference.py`, `harness.py`, and the `fake_dataset_generator.py` meta key accordingly.

## Test plan
- `uv run pytest positronic/policy/tests/test_golden_pipeline.py positronic/policy/tests/test_harness.py -p no:cacheprovider -o "addopts=" -q` → 18 passed
- `uv run pytest positronic/ pimm/ --ignore=positronic/vendors -p no:cacheprovider -o "addopts=" -q` → 530 passed, 4 skipped
- Golden verified bit-reproducible across repeated runs.


positronic/policy/tests/test_golden_pipeline.py
```diff
@@ -0,0 +1,232 @@
...
+        self.grip = pimm.ControlSystemEmitter(self)
+
+    def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock):
+        while not should_stop.value:
+            self._grip = float(self.target_grip.value)
+            self.grip.emit(self._grip)
+            yield pimm.Sleep(CONTROL_PERIOD_S)
+
+
+CAPTURED_SIGNALS = ('robot_state.ee_pose', 'robot_state.q', 'grip')
```
vertix: Please move closer to other constants

positronic/policy/harness.py
```diff
@@ -160,8 +160,14 @@ def _step(self, clock: pimm.Clock, commands_queue: deque, in_error: bool) -> Gen
             commands = self._infer(clock)
             if commands is None:
                 return False
-            if self.simulate_timeout:
-                yield pimm.Sleep(time.monotonic() - wall_start)
+            if self.simulate_inference is True:  # bool is an int subclass — check identity first
```
vertix, line 163: Make it simpler
```
if self.simulate_inferece is True:
  yield pimm.Sleep(...)
elif self.simulate_inference:
  yield ...
```

PR #406, opened 2026-05-23
Reorganize PhAIL configs by release version

vertix:
## Summary
- New `positronic.cfg.phail.v1_0` package: release manifest holding the public S3 URLs as configuronic-resolvable dataset configs (`dataset.teleoperation`, `dataset.rollouts`, `dataset.human`) plus model checkpoint paths as strings (`models.openpi/gr00t/smolvla/act`). Documents that released models were trained on the unified-task variant.
- Sim datasets (`sim_stack_cubes`, `sim_pick_place`) moved out to `positronic.cfg.ds.sim`. `positronic/cfg/ds/phail.py` deleted entirely (no back-compat).
- `utilities/release_phail.py` rewritten for the new layout: writes under `s3://positronic-public/phail/v1.0/{dataset/{teleoperation,rollouts,human},models/<vendor>}/`. New `models` subcommand copies the four trained checkpoints from `s3://checkpoints/phail_unified/...` to standardized public destinations.
- `positronic/cfg/eval.py`: added `phail_act_groot_release` source (reads `s3://inference/phail_act_groot/`) and `phail_inference_prod_v1_0` config that concats it with `phail_final` and filters to the four trained objects (`Towels`, `Wooden spoons`, `Scissors`, `Batteries`). USB-C runs (not in training data) are excluded at release time.
- Internal references updated across `workflows/nebius/`, `docs/training-workflow.md`, vendor READMEs, and `.claude/skills/remote-training/`.

Sets up the v1.0 reproducibility story: one S3 prefix per release, anyone targets `positronic.cfg.phail.v1_0` and stays version-pinned.

## Test plan
- [x] `uv run python -m py_compile positronic/cfg/phail/__init__.py positronic/cfg/phail/v1_0.py positronic/cfg/ds/sim.py utilities/release_phail.py positronic/cfg/eval.py`
- [x] `uv run python -c "from positronic.cfg.phail import v1_0; print(v1_0.dataset.rollouts); print(v1_0.models.openpi)"` resolves
- [x] `uv run python -c "from positronic.cfg import eval as e; print(e.phail_inference_prod_v1_0)"` resolves
- [x] `uv run ruff check` and `uv run ruff format --check` pass (also enforced by pre-commit hook on this commit)
- [ ] Actual release run against public S3 (`uv run python utilities/release_phail.py all`) — not run yet; requires PUBLIC_WRITE credentials and verifies end-to-end against real data


positronic/cfg/ds/sim.py
```diff
@@ -0,0 +1,20 @@
+"""Simulated datasets hosted on the public Positronic S3 bucket."""
```
vertix: Just "Positronic datasets for tasks ran in simulation"

positronic/cfg/eval.py
```diff
@@ -755,6 +757,28 @@ def _prod_predicate(ep):
...
 
+def _trained_objects_predicate(ep):
+    return ep.get('eval.object') in TRAINED_OBJECTS
+
+
+phail_act_groot_release = base_cfg.transform.override(
+    base=base_cfg.local_all.override(path='s3://inference/phail_act_groot/'),
+    transforms=[
+        Group(
+            Identity(
```
vertix: This piece is obvious duplication with `phail_inference_release` from above.
We should concat datasets and then apply transforms to them. 
Make sure there's no similar duplication here

positronic/cfg/phail/v1_0.py
```diff
@@ -0,0 +1,93 @@
...
+from positronic.dataset.transforms.episode import Derive, FromValue, Identity
+from positronic.server.positronic_server import ColumnConfig as C
+from positronic.server.positronic_server import GroupTableConfig
+from positronic.server.positronic_server import main as server_main
+from positronic.utils.logging import init_logging
+
+_ROOT = 's3://positronic-public/phail/v1.0'
+
+dataset = types.SimpleNamespace(
+    teleoperation=local_all.override(path=f'{_ROOT}/dataset/teleoperation/', profile=PUBLIC),
```
vertix: I don't like this name. Let's name it just 'teleop'

positronic/cfg/phail/v1_0.py
```diff
@@ -0,0 +1,93 @@
...
+from positronic.dataset import Episode
+from positronic.dataset.transforms.episode import Derive, FromValue, Identity
+from positronic.server.positronic_server import ColumnConfig as C
+from positronic.server.positronic_server import GroupTableConfig
+from positronic.server.positronic_server import main as server_main
+from positronic.utils.logging import init_logging
+
+_ROOT = 's3://positronic-public/phail/v1.0'
+
+dataset = types.SimpleNamespace(
```
vertix: let's call it just 'ds'. It will be shorter in configs

utilities/release_phail.py
```diff
@@ -26,13 +27,20 @@
...
-DEST_ROOT = 's3://positronic-public/datasets/phail'
+DEST_ROOT = 's3://positronic-public/phail'
 
 # Same cache path as PUBLIC (local_name='positronic-public') but with credentials for writing.
 PUBLIC_WRITE = dataclasses.replace(PUBLIC, public=False)
 
 REQUIRED_FIELDS = ['model', 'eval.object', 'eval.successful_items', 'eval.total_items']
 
+MODEL_SOURCES = {
+    'openpi': 's3://checkpoints/phail_unified/openpi/pi05_positronic_lowmem/270226-ee/',
```
vertix: The tricky part is that if we copy the full directory, we will copy not only the latest checkpoint, but all the checkpoints out there. But we need only the latest one. You may need to read the code of particular servers to understand how they deal with it

vertix: Fixed in 8f69640. Each vendor has a different checkpoint layout under `checkpoints_dir`; matched each one to its server's resolution code:

| Vendor | inner_dir | prefix | Path |
|---|---|---|---|
| openpi | — | — | `<source>/<N>/` |
| gr00t | — | `checkpoint-` | `<source>/checkpoint-<N>/` |
| smolvla | `checkpoints` | — | `<source>/checkpoints/<N>/pretrained_model/` |
| act | `checkpoints` | — | `<source>/checkpoints/<N>/pretrained_model/` |

`MODEL_SOURCES` is now `dict[str, _ModelLayout]` carrying source + `inner_dir` + `prefix` per vendor. `_copy_latest_checkpoint` calls `get_latest_checkpoint(src_root + inner_dir, prefix=...)`, then copies only `<source>/<inner_dir>/<latest>/` → `<dest>/<inner_dir>/<latest>/`. The inner structure is preserved, so a vendor server pointed at the public `models/<vendor>/` resolves the checkpoint via its normal `get_latest_checkpoint` call. Only one checkpoint per vendor lands in public S3.

PR #418, opened 2026-06-05
Make Harness name-free via `Embodiment` factory

vertix:
## Summary

Step **2b** of the PhAIL eval chain — the architectural keystone that makes the `Harness` truly **name-free**.

- **New `positronic/embodiment.py`** — the obs/command/privileged/descriptor contract as small dataclasses (`Observation`/`Command`/`Privileged`/`Embodiment`) produced by a `franka()` factory. The factory owns the canonical key names and serializers; the Harness hardcodes none.
- **Harness** drops all named device ports (`frames`/`robot_state`/`robot_commands`/…) for generic `observations`/`commands` dicts the embodiment owns. The demux is now **republish-all**: every command channel emits the waypoints the chunk carries for it; an omitted channel gets `[]` (its last-value-wins signal is overwritten, so the driver holds). This deletes the `is_recover_only` sniff, the grip-cancel special case, and the recover latency-bypass. `_home` and the descriptor come from the embodiment.
- **`wire_embodiment()`** wires the embodiment to the Harness for the inference path (`inference.py` `main`/`main_sim`); the existing `wire()` is untouched for teleop/replay. Privileged ground-truth signals are declared in the embodiment, replacing the `observers=` recording loop.
- **One serializer per observation.** `robot_state` was the only channel needing different policy vs record encodings; it's unified onto the recorded `.error` form (recording == canonical policy I/O), collapsing `Observation` to a single serializer. This pulls the step-7 reconciliation forward.
- Builds on the merged 2a (#414); also carries the `descriptor` raw-obs doc row.

**Scope:** inference path only. `DataCollectionController`/`Replay` keep their current named `wire()` wiring (the "one factory, three consumers" dedup is deferred to a later step, after bimanual).

## Behavior

Two intended changes to policy I/O, both behavior-equivalent:

1. **Recover latency.** A recover-only chunk now incurs `simulate_inference` latency in sim (the bypass is gone), so the **golden is regenerated**. The republish-all demux itself is byte-identical for the current codecs — the golden's pre-error region is byte-identical, with divergence localized to the recover point.
2. **`robot_state` serializer.** The policy obs now carries `robot_state.error` (int) instead of the raw `RobotStatus` enum, and `ErrorRecovery` reads it. `ErrorRecovery` was the enum's only consumer and `.error` carries exactly the ERROR bit it needs, so error detection is unchanged; a `RESETTING` arm now skips inference (the serializer reports not-ready) rather than being fed a partial obs.

Minor: `static.json` key *insertion order* shifts for sim recordings (`simulation.mujoco_model_path` moves first→last) — same keys, same values; cosmetic. The real-hardware path is order-preserving.

## Test plan

- `uv run --locked pytest positronic/policy/tests/ positronic/tests/test_inference_integration.py positronic/tests/test_data_collection.py --no-cov` — 114 passed.
- Golden regenerated via `GOLDEN=1 …` and re-asserted green.
- End-to-end `main_sim` integration test exercises `franka` → `wire_embodiment` → Harness → recorded dataset (records `robot_commands.pose`, `target_grip`, `image.*`, privileged `sim_state.*`).


vertix:
@codex review

positronic/policy/harness.py
```diff
@@ -366,12 +381,7 @@ def _step(self, clock: pimm.Clock) -> Generator[pimm.Sleep, None, None]:
...
-        # Recover-only output (from ``ErrorRecovery``) bypasses inference latency
-        # simulation — it's an emergency emit, not a model-driven chunk.
-        is_recover_only = len(actions) == 1 and isinstance(actions[0].get('robot_command'), roboarm.command.Recover)
-        if is_recover_only:
-            delay = 0.0
-        elif self.simulate_inference is True:  # bool is an int subclass — check identity first
+        if self.simulate_inference is True:  # bool is an int subclass — check identity first
             delay = time.monotonic() - wall_start
         elif self.simulate_inference:
             delay = float(self.simulate_inference)
```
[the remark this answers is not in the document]
vertix: Intentional and sim-only — declining.

`simulate_inference` is only ever set on the sim path (`main_sim`); the real path (`main`) never passes it, so on real hardware `delay = 0` and a `Recover` is still emitted immediately with no added latency. Deleting the recover latency-bypass is a deliberate part of this PR: `simulate_inference` is a sim-only simulation of model latency, and sim is safe, so a recover chunk incurring the same simulated delay is acceptable (the golden was regenerated for exactly this one timing shift). Removing the `is_recover_only` special-case is the point — the demux has no per-command branching anymore.

positronic/policy/harness.py
```diff
@@ -323,36 +326,48 @@ def _handle_directive(
...
+        it; a channel an action omits gets ``[]`` — overwriting its last-value-wins
+        signal, so the driver holds. An empty ``actions`` therefore cancels every
+        channel. No channel is special-cased: arm and gripper are treated alike.
+        """
+        for name, emitter in self.commands.items():
+            # Wrappers do action-timing math in float seconds (codecs are fps-based);
+            # clients on every pimm channel (driver TrajectoryPlayer, dataset writer)
+            # expect ns. This is the single explicit seconds->ns seam.
+            traj = [(int(a['timestamp'] * 1e9), a[name]) for a in actions if name in a]
+            emitter.emit(traj)
```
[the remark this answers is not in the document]
vertix: Good catch — this is a real divergence, but it's confined to chunks that omit a channel, which no in-repo codec produces (every action codec emits both `robot_command` and `target_grip`). The one grip-less chunk that actually occurs — `ErrorRecovery`'s `Recover` — *should* cancel the gripper buffer (the prior behavior, guarded by `test_recovery_cancels_gripper_buffer`), and republish-all preserves that.

The remaining case you flag — a remote/legacy policy returning a `robot_command`-only chunk — is the asymmetric-rate-channel scenario the design explicitly defers to a later step (see `## 12 — Driver uniformity: gripper as a command stream` in `docs/phail-eval-mvp.md`). That step makes `TrajectoryPlayer.set([])` a genuine no-op (keep playing) and adds an explicit `Hold`, so an omitted channel holds *without* clearing the in-flight buffer or truncating the recorder tail. Until then, emitting `[]` to an omitted channel (the driver holds its last commanded value) is the intended republish-all semantics, and a clean fix here can't distinguish "hold this channel" from "cancel this channel" without re-introducing the per-channel special-casing this PR removed. Deferring rather than half-fixing.

positronic/embodiment.py
```diff
@@ -0,0 +1,109 @@
...
+
+# Embodiment-level static meta: how recorded signals map to the canonical robot fields.
+ROBOT_STATIC_META = {'joint_signal': 'robot_state.q', 'pose_signals': ['robot_state.ee_pose', 'robot_commands.pose']}
+
+
+@dataclass
+class Observation:
+    """A policy-facing signal source plus its two serializers.
+
+    ``to_policy`` builds the observation entries fed to the policy; ``to_record``
```
vertix: Should we add TODO, that this class should go away when we generalise serialisation so that it is by-type?

positronic/embodiment.py
```diff
@@ -0,0 +1,109 @@
...
+    builds the entries written to the dataset. They differ today (the obs side keeps
+    the raw ``RobotStatus`` that ``ErrorRecovery`` matches on); step 7 unifies them.
+    ``None`` on either side passes the device value through unchanged.
+    """
+
+    source: pimm.SignalEmitter
+    to_policy: Serializer | None
```
vertix: Pls explain here why we have 2 serialisers. Also, we should simplify the case when both serialisers are the same

positronic/embodiment.py
```diff
@@ -0,0 +1,109 @@
...
+    """A policy action channel: where its waypoints go and how it homes/records.
+
+    ``home`` is the value emitted to send this channel to its safe state. ``record_name``
+    is the dataset signal name (it may differ from the action key, e.g. the ``robot_command``
+    action records under ``robot_commands``). ``to_record`` serializes the channel's values.
+    """
+
+    dest: pimm.SignalReceiver
+    home: Any
+    record_name: str
```
vertix: Why do we need a seperate record_name? Why can't we just use the key?

positronic/embodiment.py
```diff
@@ -0,0 +1,109 @@
...
+    Backed by 1 or N device control systems (not fused) — what satisfies the
+    contract is an implementation detail. The Harness is name-free: it reads the
+    serializers and command spec from here and never hardcodes a canonical key.
+    """
+
+    descriptor: str
+    observations: dict[str, Observation]
+    commands: dict[str, Command]
+    privileged: dict[str, Privileged]
+    static_meta: dict[str, Any]
```
vertix, line 59: Should it be the factory or a constant? Any trade-offs?

positronic/embodiment.py
```diff
@@ -0,0 +1,109 @@
...
+    meta_source: pimm.SignalEmitter | None
+
+    @property
+    def home(self) -> dict[str, Any]:
+        """The home action: ``{command_name: home_value}`` for every channel."""
+        return {name: cmd.home for name, cmd in self.commands.items()}
+
+
+def franka(
```
vertix: Wait, should not we also have simulator? Isn't it the whole point?

positronic/embodiment.py
vertix: Done in ff4182c — added a TODO on `Observation` noting that both serializers (and most of the class's reason to exist) collapse once serialization is type-owned (steps 7-9), where the value's domain type owns its policy- and dataset-side encoding, so the channel won't carry serializers at all.

positronic/embodiment.py
vertix: Done in ff4182c.

**Why two serializers:** the policy and the dataset can need *different* encodings of the same device value. Today only `robot_state` actually differs — the obs side keeps the raw `RobotStatus` that `ErrorRecovery` matches on, while the record side drops `RESETTING` and emits `.error`. Everywhere else they coincide. Expanded the docstring to say this.

**Simplified the same-serializer case:** `to_record` now defaults to `to_policy`, so channels where the two coincide stop repeating themselves — `franka` now writes `Observation(gripper.grip, None)` and `Observation(emitter, Serializers.camera_images)`, with `robot_state` the only channel passing both explicitly.

positronic/embodiment.py
vertix: The recorded signal is `robot_commands` (plural) while the action key the policy emits — and the demux channel name — is `robot_command` (singular). The plural name is load-bearing: training codecs key off it (`IKJointsAction(tgt_ee_pose_key='robot_commands.pose')`), `ROBOT_STATIC_META['pose_signals']` lists `robot_commands.pose`, and existing on-disk datasets already use it. `record_name` lets the action key stay singular (matching the channel) while the dataset keeps the established plural name.

We *could* collapse them by renaming the recorded signal to `robot_command` everywhere, but that's a cross-cutting rename (codecs + static meta + on-disk datasets) I'd keep out of this PR. Note `target_grip`'s `record_name` already equals its key, so the field only does real work for the arm channel — happy to drop it and special-case just the arm rename instead if you'd prefer that trade.

positronic/embodiment.py
vertix: Constant dict today. The only genuinely dynamic meta — `urdf`, `joint_names`, `control_frame` — already flows separately through `meta_source` → the harness's `robot_meta_in` receiver (the device emits it at runtime), so the embodiment's `static_meta` only needs the truly static mapping (`ROBOT_STATIC_META`) plus run config (`simulation.mujoco_model_path`).

Trade-off: a constant is simplest and eager; a `Callable[[], dict]` factory would buy lazy / per-episode computation, but we have no case that needs it (the dynamic path already exists via `robot_meta_in`). I'd promote it to a factory only when an embodiment needs episode-invariant meta that isn't known at construction time — cheap to change later since the harness just consumes `embodiment.static_meta`.

positronic/embodiment.py
vertix: `franka()` *is* the sim factory — and the real one. It's parametrized over the arm/gripper control systems, which are `MujocoFranka`/`MujocoGripper` in sim (see `inference.main_sim`) and the DROID drivers on hardware; that's what the docstring's "shared by the sim, real, and golden paths" means. The simulator enters two ways, both via the caller rather than the factory: (1) you pass the Mujoco device CSs as `robot_arm`/`gripper`, and (2) sim ground truth comes in as `privileged={'sim_state': sim.observations['sim_state']}`.

Keeping the factory ignorant of `MujocoSim` is deliberate — it only ever sees signal sources, so one builder serves sim and real without a `mujoco_` fork. If you meant a convenience `mujoco_franka(sim, ...)` wrapper that wires those up for you, that's easy to add — say the word.

positronic/embodiment.py
vertix: Implemented in 39b8f1a. `robot_state` was the only channel forcing two serializers, so it now uses the single `robot_state` serializer for both the policy and the recording (the `.error` form). `ErrorRecovery` — the raw enum's only consumer — now reads `robot_state.error`, and a `RESETTING` arm makes the serializer report not-ready so `_build_obs` skips inference instead of feeding a partial obs.

With that, no channel needs two serializers, so `Observation` collapses to a single `serializer` field and the `_SAME_AS_POLICY` sentinel is gone. (This pulls the step-7 reconciliation forward.)

positronic/embodiment.py
```diff
@@ -0,0 +1,113 @@
...
+    source: pimm.SignalEmitter
+    to_record: Serializer | None
+
+
+@dataclass
+class Embodiment:
+    """The signal-dict contract the Harness drives, produced by a factory.
+
+    Backed by 1 or N device control systems (not fused) — what satisfies the
+    contract is an implementation detail. The Harness is name-free: it reads the
```
vertix: Nobody reading this piece will ever know that harness was ever non name-free. Update docstrings EVERYWHERE so they don't have archeology. For the new reader

positronic/wire.py
```diff
@@ -57,3 +58,43 @@ def wire(
...
+    for name, cmd in embodiment.commands.items():
+        world.connect(harness.commands[name], cmd.dest)
+    if embodiment.meta_source is not None:
+        world.connect(embodiment.meta_source, harness.robot_meta_in)
+
+    ds_agent = None
+    if dataset_writer is not None:
+        ds_agent = DsWriterAgent(dataset_writer, time_mode=time_mode)
+        for name, obs in embodiment.observations.items():
+            ds_agent.add_signal(name, obs.serializer)
```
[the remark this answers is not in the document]
vertix: Fixed in 3428eaf. `wire_embodiment` now raises if an observation serializer is a `StatefulSerializer` instance, so the shared-object foot-gun fails loud at wire time instead of silently corrupting the policy path when the writer resets state at episode boundaries.

It's inert today — `robot_state`/`camera_images` are pure functions, which `add_signal` wraps in a fresh `_PureSerializer` per consumer (no aliasing). The guard fires only the moment someone introduces a stateful observation serializer instance, where the real fix is per-caller instances (type-owned serialization, the steps 8-9 TODO on `Observation`); per-caller divergence there is acceptable and intended.

Note it lives only here, not in `add_signal`: `wire()`'s teleop path uses a stateful serializer (`TrajectoryOverrideSerializer`) safely as the sole consumer, so the constraint is specific to the obs serializer being shared with the harness.

positronic/embodiment.py
vertix: Done in 2250991. Rewrote the docstrings I added to describe what the contract *is* for a fresh reader, dropping the framing that only parses if you know the harness used to be name-bound: `name-free`, `never hardcodes a canonical key`, `no arm-vs-gripper knowledge / no per-command special-casing`, `treated alike`. Also trimmed the redundant comments — e.g. the `# Embodiment descriptor … passed to the policy every call` line you flagged.

Scoped to the docstrings/comments this PR introduced; I left pre-existing docstrings untouched (not part of the diff). If you want the de-archeology to reach beyond the changed lines, point me at the spots and I'll extend it.

positronic/cfg/embodiment.py
```diff
@@ -0,0 +1,112 @@
...
+    cameras={
+        'image.wrist': positronic.cfg.hardware.camera.zed_m.override(
+            view='left', resolution='hd720', fps=30, image_enhancement=True
+        ),
+        'image.exterior': positronic.cfg.hardware.camera.zed_2i.override(
+            view='left', resolution='hd720', fps=30, image_enhancement=True
+        ),
+    },
+)
+def droid(robot_arm, gripper, cameras):
```
vertix, line 26: Configuronic way is the different.
* We define one general config (here 'franka', but franka is a bad name because one can think about real franka, and it is not very descriptive TBH, it can be any arm actually), and then define `droid` and `sim` as overrides.
* Please also recommend me how to change the documentation in configuronic, so that the agents like you figure this recommendation out on your own. Analyse what code did you read when creating this

positronic/cfg/embodiment.py
```diff
@@ -0,0 +1,112 @@
...
+    mujoco_model_path=package_assets_path('assets/mujoco/franka_table.xml'),
+    loaders=positronic.cfg.simulator.stack_cubes_loaders,
+    camera_fps=15,
+    # We use 3 cameras not because we need it, but because Mujoco does not render
+    # the second image when using only 2 cameras
+    camera_dict={'image.wrist': 'handcam_left_ph', 'image.exterior': 'back_view_ph', 'image.agent_view': 'agentview'},
+    # Full sim state is the privileged ground truth; scoring is computed downstream.
+    observers={'sim_state': FullSimState()},
+)
+def sim(mujoco_model_path, loaders, camera_fps, camera_dict, observers):
```
vertix: This is a bad name, because there are potentially multiple different sims.

positronic/inference.py
```diff
@@ -122,113 +116,49 @@ def _connect_ds_command(world, harness, ds_agent, policy):
...
-        if ds_agent is not None:
-            for observer_name in observers.keys():
-                ds_agent.add_signal(observer_name)
-                world.connect(sim.observations[observer_name], ds_agent.inputs[observer_name])
-        _connect_ds_command(world, harness, ds_agent, policy)
+    with writer_cm as dataset_writer, pimm.World(virtual_time=embodiment.simulated) as world:
+        ds_agent = wire.wire_embodiment(world, harness, embodiment, dataset_writer, time_mode)
+        if gui is not None:
+            for name, obs in embodiment.observations.items():
+                if name.startswith('image.'):
```
vertix, line 144: Please add a comment (HACK or TODO) that this is a hard binding that needs to be resolved somehow

positronic/inference.py
```diff
@@ -237,16 +167,18 @@ def main_sim(
...
-        'sim': main_sim_cfg,
-        'real': droid_setup,
-        'phail': droid_setup.override(
-            policy=policy_cfg.phail_multiple, driver=eval_ui, **{'driver.ui_scale': 3, 'robot_arm.collision_coeff': 2.0}
-        ),
-        'sim_pnp': main_sim_cfg.override(
-            loaders=positronic.cfg.simulator.multi_tote_loaders,
-            observers={},
-            **{'driver.task': 'Pick up objects from the red tote and place them in the green tote.'},
+        'run': run_cfg,
```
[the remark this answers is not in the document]
vertix: Fixed in 9e97173: re-added `'real': run_cfg` as a back-compat alias, so `positronic-inference real ...` still routes to the hardware (droid) path. `run` and `real` are the same config (both default `embodiment=droid`); `run` is the new generic name that will take a bare `--embodiment` once configuronic supports a default command.

positronic/inference.py
vertix: Done in 9e97173: added a HACK/TODO noting the GUI cameras are bound to observations by the `image.` name-prefix convention, and that this should be replaced by an explicit camera declaration on the embodiment rather than a string prefix.

positronic/policy/harness.py
```diff
@@ -130,7 +131,7 @@ def __init__(self, inner: Session, clock: pimm.Clock):
 
         def __call__(self, obs):
             was_ok = not self._in_error
-            self._in_error = obs['robot_state.status'] == roboarm.RobotStatus.ERROR
+            self._in_error = obs['robot_state.error'] == 1
```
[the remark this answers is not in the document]
vertix: Valid for the multi-embodiment future, but out of scope for this PR. We ship a single-arm Franka here, where `robot_state.error` and the `robot_command` channel exist by construction, so `default_wrappers` is correct.

The deeper fix — parameterizing `ErrorRecovery` off the embodiment (the error-observation key and the recovery command channel) so non-Franka setups can keep the default wrappers — belongs with the broader recovery/serialization generalization (the type-owned serialization + bimanual steps), not this name-free wiring change. Leaving open to track.

positronic/cfg/embodiment.py
vertix: Done in f57e9a6 — went simpler than the override-a-general-config idiom: dropped the shared `franka` builder entirely. `droid`, `mujoco_franka`, and `mujoco_franka_pnp` are now standalone configs that each assemble their `Embodiment` inline, no shared function.

On the idiom recommendation you asked for: I dug into why a single general-config + overrides doesn't cleanly cover the sim case — its three devices (arm/gripper/cameras) share one `MujocoSim`, and configuronic's `instantiate` builds every `Config` reference independently (no memoization), so they can't share through separate sub-configs. I filed that gotcha plus the missing "variants via `.override()`" docs as configuronic#29 (and the `''` default-command request as #28), so future agents learn the idiom from the docs.

positronic/cfg/embodiment.py
vertix: Renamed in f57e9a6: `sim` → `mujoco_franka` (names the platform + robot, since there can be many sims). The multi-tote pick-and-place variant is `mujoco_franka_pnp`.

positronic/embodiment.py
vertix: Resolved by the restructure: as of f57e9a6 the simulator is a first-class embodiment config — `mujoco_franka` (and `mujoco_franka_pnp`) in `cfg/embodiment.py`, selected via `--embodiment` / the `sim` command — rather than being implicit in a shared factory.

positronic/embodiment.py
vertix: Fixed in 060c864: removed `Command.record_name`. Each channel records under its own key now, and the recorded arm command went `robot_commands.pose` -> `robot_command.pose` (singular, consistent with `robot_state`). Existing S3 datasets are migrated with a lazy zero-copy alias in `cfg/ds`.

positronic/policy/harness.py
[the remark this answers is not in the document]
vertix: Intended behavior. This is the uniform republish-all demux at the core of the name-free harness: every channel is emitted from each chunk, and a channel the chunk omits gets `[]`. Special-casing omitted channels would reintroduce the per-channel logic the refactor removed. In practice our codecs always emit `robot_command` and `target_grip` together, so a real chunk never drops the gripper. Resolving as won't-fix.

positronic/replay_record.py
```diff
@@ -51,7 +51,7 @@ def __init__(self):
 
     @staticmethod
     def _commands_from_episode(episode: Episode) -> Any:
-        pose = episode['robot_commands.pose']
+        pose = episode['robot_command.pose']
```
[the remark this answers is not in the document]
vertix: Fixed in 985bccc: `RestoreCommand._commands_from_episode` falls back to the legacy `robot_commands.pose` key, so the default replay path handles pre-rename recordings without the compat transform.

positronic/policy/harness.py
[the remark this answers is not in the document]
vertix: Acknowledged and documented in 7bb3308 with a TODO on `ErrorRecovery`: the wrapper hard-codes the `robot_state.error` observation and `robot_command` channel (Franka `Recover`), so non-Franka embodiments must disable `default_wrappers` for now. Deferring the generalization until a real second embodiment exists to design it against — building it now would be a speculative abstraction. Resolving against the TODO.

positronic/embodiment.py
```diff
@@ -0,0 +1,74 @@
...
+    serializes the channel's values, recorded under the channel's own key.
+    """
+
+    dest: pimm.SignalReceiver
+    home: Any
+    serializer: Serializer | None
+
+
+@dataclass
+class Privileged:
```
vertix: Do we really need this class, as it is duplicate of Observation above.

vertix: Agreed — dropped in b381c51. `Privileged` was structurally identical to `Observation` (`source` + `serializer`) with no `isinstance` dispatch anywhere; the "recorded but never fed to the policy" distinction is carried by the `privileged` dict (kept out of `observations`), not by the type. The dict is now `dict[str, Observation]` with a one-line comment noting the ground-truth semantic.

PR #419, opened 2026-06-05
Register `PUBLIC` pos3 profile for anonymous public-bucket access

vertix:
## Summary

Lets a reader with **no AWS credentials** serve the public phail models directly, by making the `s3://PUBLIC@positronic-public/...` URL form resolve to anonymous (unsigned) access everywhere.

- **`positronic/utils/__init__.py`** — define the public S3 `Profile` here (a low-level module every S3 entrypoint already imports, at zero added import cost) and `register_profile('PUBLIC', …)` from it. This is the one place reachable by inference servers, which take a bare `--checkpoints_dir` string and never import `cfg.ds`. `positronic/__init__.py` stays minimal, so pure-compute imports don't pull in boto3.
- **`positronic/cfg/ds/__init__.py`** — re-export `PUBLIC` from `utils` instead of holding a duplicate `Profile` (single source of truth; utils never imports cfg, so no cycle). `cfg.phail`, `cfg.ds.sim`, and the lerobot server keep importing it from `cfg.ds` unchanged.
- **`positronic/cfg/phail/v1_0.py`** — point the v1.0 **model** URLs at the `s3://PUBLIC@…` form so they resolve anonymously on their own (datasets keep `profile=PUBLIC`, which they pass explicitly).

Every pos3 op (`download`/`ls`/checkpoint listing) routes through `_effective_profile(profile, remote)`, where the URL's `@PUBLIC` wins — so the servers' existing bare `pos3.download` / `pos3.ls` need no change.

## Test plan

New offline regression suite `positronic/utils/tests/test_profiles.py` (5 tests):
- registration + fields, re-export identity, phail model URLs use the `PUBLIC@` form;
- `public=True` yields an **UNSIGNED** client (the no-creds property, asserted without network);
- a fresh interpreter importing only `utils.checkpoints` (what servers use, never `cfg.ds`) must already resolve `PUBLIC` — guards the design against the registration drifting to a module servers don't import.

Manually verified anonymous access with **all** AWS credential sources scrubbed in a fresh interpreter:
- bare `s3://positronic-public/...` (signed) → `NoCredentialsError` (negative control);
- `s3://PUBLIC@positronic-public/phail/v1.0/models/gr00t/` → lists `checkpoint-150000`.

`pytest positronic/cfg/tests positronic/utils/tests --no-cov` → 24 passed.

### DIFF
--- a/positronic/cfg/ds/__init__.py
+++ b/positronic/cfg/ds/__init__.py
@@ -12,10 +12,7 @@ from positronic.dataset.dataset import ConcatDataset, Dataset, FilterDataset
 from positronic.dataset.local_dataset import LocalDataset, load_all_datasets
 from positronic.dataset.transforms import TransformedDataset
 from positronic.dataset.transforms.episode import EpisodeTransform, Group
-
-PUBLIC = Profile(
-    local_name='positronic-public', endpoint='https://storage.eu-north1.nebius.cloud', public=True, region='eu-north1'
-)
+from positronic.utils import PUBLIC
 
 
 @cfn.config()
--- a/positronic/cfg/ds/__init__.py
+++ b/positronic/cfg/ds/__init__.py
@@ -8,11 +8,11 @@ import configuronic as cfn
 import pos3
 from pos3 import Profile
 
+import positronic.utils  # noqa: F401  -- registers the 'PUBLIC' S3 profile that these loaders' s3://PUBLIC@ URLs use
 from positronic.dataset.dataset import ConcatDataset, Dataset, FilterDataset
 from positronic.dataset.local_dataset import LocalDataset, load_all_datasets
 from positronic.dataset.transforms import TransformedDataset
 from positronic.dataset.transforms.episode import EpisodeTransform, Group
-from positronic.utils import PUBLIC
 
 
 @cfn.config()
--- a/positronic/cfg/phail/v1_0.py
+++ b/positronic/cfg/phail/v1_0.py
@@ -22,6 +22,10 @@ from positronic.server.positronic_server import main as server_main
 from positronic.utils.logging import init_logging
 
 _ROOT = 's3://positronic-public/phail/v1.0'
+# Model URLs carry the `PUBLIC@` profile selector so they resolve to anonymous access on
+# their own — a no-creds reader can pass them straight to an inference server's
+# `--checkpoints_dir`, which never sees the PUBLIC object the dataset configs pass explicitly.
+_MODELS_ROOT = f'{_ROOT.replace("s3://", "s3://PUBLIC@")}/models'
 
 ds = types.SimpleNamespace(
     teleop=local_all.override(path=f'{_ROOT}/dataset/teleoperation/', profile=PUBLIC),
@@ -30,10 +34,10 @@ ds = types.SimpleNamespace(
 )
 
 models = types.SimpleNamespace(
-    openpi=f'{_ROOT}/models/openpi/',
-    gr00t=f'{_ROOT}/models/gr00t/',
-    smolvla=f'{_ROOT}/models/smolvla/',
-    act=f'{_ROOT}/models/act/',
+    openpi=f'{_MODELS_ROOT}/openpi/',
+    gr00t=f'{_MODELS_ROOT}/gr00t/',
+    smolvla=f'{_MODELS_ROOT}/smolvla/',
+    act=f'{_MODELS_ROOT}/act/',
 )
 
 UNIFIED_TASK = 'Pick all the items one by one from transparent tote and place them into the large grey tote.'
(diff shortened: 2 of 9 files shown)


positronic/cfg/phail/v1_0.py
```diff
@@ -22,6 +22,10 @@
 from positronic.utils.logging import init_logging
 
 _ROOT = 's3://positronic-public/phail/v1.0'
+# Model URLs carry the `PUBLIC@` profile selector so they resolve to anonymous access on
+# their own — a no-creds reader can pass them straight to an inference server's
+# `--checkpoints_dir`, which never sees the PUBLIC object the dataset configs pass explicitly.
+_MODELS_ROOT = f'{_ROOT.replace("s3://", "s3://PUBLIC@")}/models'
```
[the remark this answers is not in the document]
vertix: Fixed in b5ce108: bumped the floor to `pos3>=0.3.1` and re-locked (resolved version stays 0.3.1). The `s3://<profile>@bucket` URL selector is a 0.3.x feature, and the repo already baselines at 0.3.1 (#411), so supporting below it serves no one.

vertix: Why just not to print the resolved path, instead of this replacements. It's not worth it

positronic/utils/__init__.py
```diff
@@ -9,13 +9,25 @@
...
 from positronic import __file__ as pkg_init_file
 from positronic.utils.checkpoints import get_latest_checkpoint, list_checkpoints
 from positronic.utils.frozen_dict import frozen_keys_dict, frozen_view
 from positronic.utils.git import get_git_diff, get_git_state
 
+# Positronic's public S3 bucket. Defined here (a low-level module every S3 entrypoint
+# imports) and registered under the name 'PUBLIC' so the `s3://PUBLIC@positronic-public/...`
+# URL form resolves to anonymous access everywhere — including inference servers that take a
+# bare `--checkpoints_dir` string and never see the PUBLIC object.
+PUBLIC = pos3.Profile(
```
vertix, line 25: Do we really need this profile now? Should we migrate to PUBLIC@ everywhere?

pyproject.toml
```diff
@@ -32,7 +32,7 @@ dependencies = [
     "opencv-python-headless",
     # openpi-client: optional, see [project.optional-dependencies] openpi
     "plotly>=6.5.0",
-    "pos3>=0.2.0",
+    "pos3>=0.3.1",  # s3://<profile>@bucket URL selectors (anonymous public access) need 0.3.x
```
vertix: Please remove the comment.

positronic/cfg/phail/v1_0.py
vertix: Fixed in 22c4b24: folded `PUBLIC@` straight into `_ROOT` (`s3://PUBLIC@positronic-public/phail/v1.0`), so the `.replace()` / `_MODELS_ROOT` is gone — datasets and models both build off it directly.

positronic/utils/__init__.py
vertix: Done in 22c4b24: dropped the `PUBLIC` Profile object and migrated every `profile=PUBLIC` site to the `s3://PUBLIC@...` URL form (phail datasets + models, `cfg.ds.sim`, lerobot `demo`). `utils` now just `register_profile('PUBLIC', ...)`.

One wrinkle worth flagging: dropping the re-export meant `cfg.ds` no longer imported `utils`, so `sim.py` (which only imports `cfg.ds`) would lose the registration trigger. `cfg/ds/__init__.py` now imports `positronic.utils` for that side effect, guarded by a new subprocess test (`test_registration_reaches_cfg_ds_consumers`).

pyproject.toml
vertix: Removed in 22c4b24.

positronic/cfg/ds/__init__.py
```diff
@@ -8,15 +8,12 @@
 import pos3
 from pos3 import Profile
 
+import positronic.utils  # noqa: F401  -- registers the 'PUBLIC' S3 profile that these loaders' s3://PUBLIC@ URLs use
```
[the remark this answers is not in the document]
vertix: Good catch — fixed in f4a3ec5, and you're right that the object is still needed. `release_phail` derives a write-credentialed profile (`PUBLIC_WRITE = dataclasses.replace(PUBLIC, public=False)`), which a `PUBLIC@` URL can't express (you can't write anonymously).

I took the second of your two options rather than re-exporting from `cfg.ds`: the canonical `PUBLIC` Profile object now lives in `positronic.utils` (where it's also registered), and `release_phail` imports it from there. That keeps `cfg.ds`'s surface URL-only (the migration this PR is about) while preserving the object for the one caller that genuinely needs it as a value. Verified `utilities/release_phail.py` imports and `PUBLIC_WRITE` derives (`public=False`).

PR #420, opened 2026-06-09
Add `positronic eval run` with first-class `Task` + `Eval`

vertix:
## Summary

Step 3 of the PhAIL eval chain: introduce a first-class **Task** abstraction, realize **eval = embodiment + task**, and put the researcher CLI (`positronic eval run`) on top.

- **`Task` + `Eval`** (`positronic/task.py`): `Task(instruction, privileged)` and `Eval(embodiment, task)` — plain dataclasses, no ABC. An eval is a single config that **holds the sim** and builds both the (pure-robot) embodiment and the task from it.
- **Privileged off the embodiment, instruction off the driver.** `Embodiment.privileged` is removed; the Harness injects the task instruction on the `RUN` directive (`instruction` param), and `wire_embodiment` records the task's privileged signals via a `privileged=` param. `mujoco_franka` becomes a pure-robot builder over a shared `MujocoSim`.
- **`positronic` parent command** with an `eval` group and a `run` verb. Evals are selected by **lazy relative reference** — `positronic eval run --eval=.sim.positronic.stack_cubes` — resolved against a sentinel base in `positronic.eval`, importing only the selected eval. No central catalog dict.
- **Migration:** today's `positronic-inference sim`/`sim_pnp` become eval configs `sim.positronic.stack_cubes` / `sim.positronic.multi_tote`. `inference.main` gains `task` + `wrap` (passthrough). Docs (vendor READMEs, `docs/inference.md`, etc.) and `utilities/validate_server.py` updated to the new command.

`list` is intentionally deferred to its own follow-up PR (it needs namespace discovery).

## configuronic 0.5.0

`positronic eval run` uses configuronic's **nested command tree** (`cfn.cli({'eval': {'run': ...}})`), which ships in configuronic **0.5.0**. This PR pins `configuronic>=0.5.0` and relocks (`uv.lock` updated 0.4.0 → 0.5.0). Verified `positronic eval run --eval=.sim.positronic.stack_cubes` resolves end-to-end through the nested tree.

## Test plan

- `uv run --locked pytest --no-cov positronic/policy/tests/ positronic/tests/test_inference_integration.py` — **111 passed** on configuronic 0.5.0 (incl. the reworked integration test exercising the full eval→task→harness path: privileged `sim_state` recording + harness-injected instruction).
- `uv run --locked ruff check` + `ruff format --check` on all changed files — clean.
- `positronic eval run --eval=.sim.positronic.stack_cubes --help` resolves through the nested command tree to `_mujoco_franka_eval` with the right instruction/loaders/observers (and `.multi_tote` likewise).
- Verified `utilities/validate_server.py` now builds `positronic eval run --eval=.sim.positronic.stack_cubes …`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


positronic/cli.py
```diff
@@ -0,0 +1,17 @@
...
+from positronic.eval import run_cfg
+from positronic.utils.logging import init_logging
+
+
+# The `positronic` parent command. Only `eval` lives under it today; `inference`/`server`
+# fold in as sibling groups in a later, mechanical follow-up.
+@pos3.with_mirror()
+def _internal_main():
+    init_logging()
+    cfn.cli({'eval': {'run': run_cfg}})
```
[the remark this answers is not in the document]
vertix: Correct, and this is the pending **configuronic 0.5.0** dependency called out in the PR description. `positronic eval run` uses configuronic's nested command tree (`cfn.cli({'eval': {'run': ...}})`), which ships in configuronic **0.5.0**; the locked **0.4.0** only dispatches a single command level (`_cli_multiple_commands`), so this `AttributeError` is expected until `configuronic>=0.5.0` is pinned.

That release isn't on PyPI yet (latest is 0.4.0), so this PR deliberately doesn't bump the pin. The `configuronic>=0.5.0` pin + `uv lock` will land as a follow-up commit before merge, at which point `positronic eval run` (and `validate_server.py`) resolve through the nested tree. Eval *selection* (relative-ref resolution, `--eval=.sim.positronic.stack_cubes`) is a `Config` feature already present in 0.4.0 and is verified working.

Leaving this open until the 0.5.0 pin lands.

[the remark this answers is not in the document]
vertix: Resolved in fb531ed: pinned `configuronic>=0.5.0` and relocked (`uv.lock` 0.4.0 → 0.5.0). configuronic 0.5.0 (with the nested command tree) is now published to PyPI, so `positronic eval run --eval=.sim.positronic.stack_cubes` dispatches end-to-end — verified via `--help` resolving through `{'eval': {'run': run_cfg}}` to `_mujoco_franka_eval` with the right instruction/loaders/observers, and the 111-test suite passes on 0.5.0.

positronic/cli.py
```diff
@@ -0,0 +1,17 @@
+import configuronic as cfn
+import pos3
+
+from positronic.eval import run_cfg
+from positronic.utils.logging import init_logging
+
+
+# The `positronic` parent command. Only `eval` lives under it today; `inference`/`server`
```
vertix: The comment mentions the future changes. This is not helpful, we should either leave a TODO, or (preferably in our case) say something about the current state of affairs.
If there's nothing to say beyond the code, just stay silent

positronic/task.py
vertix, line 1: We don't need to many files. I prefer to merge task.py with embodiment.py into something more broader like eval.py. No need to introduce extra instances with low value

positronic/eval/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def _eval_required():
+    raise SystemExit('--eval is required, e.g. --eval=.sim.positronic.stack_cubes')
+
+
+# Sentinel default defined in this package so its base path is ``positronic.eval``:
+# ``--eval=.sim.positronic.stack_cubes`` resolves to ``positronic.eval.sim.positronic.stack_cubes``,
+# lazily (only the selected eval is imported) — there is no central catalog to keep in sync.
+EVAL_SENTINEL = cfn.Config(_eval_required)
```
vertix: This should be done as function decorator. The function can be names in uppercase.
Same below. Please update your memory to strongly prefer the decorator

positronic/eval/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+        driver=driver,
+        output_dir=output_dir,
+        simulate_inference=simulate_inference,
+        wrap=wrap,
+    )
+
+
+run_cfg = cfn.Config(
+    _run,
+    eval=EVAL_SENTINEL,
```
vertix: we should call it placeholder, as below for policy

positronic/cfg/embodiment.py
vertix, line 24: Why we don't see the configuration for droid pick-and-place task?
I'd love to see it in this PR, we have a general pick and place task, etc.

positronic/task.py
```diff
@@ -0,0 +1,29 @@
...
+class Task:
+    """The scenario layered on an embodiment: the policy-facing instruction plus
+    the privileged ground-truth signals to record.
+
+    ``instruction`` is the language goal sent to the policy on every call. ``privileged``
+    maps a record key to the ground-truth source to capture (the sim's full ``save_state``,
+    a real scale) — recorded but never fed to the policy.
+    """
+
+    instruction: str
```
vertix: I think we discussed this with you in our document, but we should have
a) seed
b) timeout 
somewhere in our eval structures. May be in task, may be in eval, let's discuss and make it.

positronic/policy/harness.py
```diff
@@ -285,6 +287,9 @@ def _handle_directive(
                     self._home(clock)
                     yield pimm.Yield()
                 self.context = directive.payload or {}
+                # The task's instruction is authoritative for the run; drivers no longer carry it.
```
vertix: All the documentation should not have any references to past or future (except TODOs). No archeology, we assume that the document is always read with the fresh eyes.

No need to explain what does not require explanation on top of the code

positronic/cli.py
vertix: Fixed in 4151a63: dropped the comment entirely — nothing to say beyond the code here.

positronic/task.py
vertix: Fixed in 4151a63: `task.py` + `embodiment.py` merged into `positronic/eval.py` — one leaf module with `Observation`/`Command`/`Embodiment`/`Task`/`Eval`. The runner config moved to `positronic/cfg/eval/` (configs belong in `cfg/`), and the analysis configs formerly at `cfg/eval.py` were renamed to `cfg/analysis.py` to free the namespace.

positronic/eval/__init__.py
vertix: Fixed in 4151a63: both configs use the `@cfn.config` decorator form. Memory updated to strongly prefer the decorator.

positronic/eval/__init__.py
vertix: Fixed in 4151a63: renamed to `placeholder`, mirroring `cfg/policy.py`.

positronic/cfg/embodiment.py
vertix: Deferred to step 6 (`real.droid.pick_place`) per our discussion. Registering it properly needs the attended-driver refit we designed in the doc: `EvalUI` wraps the `Eval` (defaults from the task, may redefine instruction/timeout per trial), stays the path's single directive emitter (a hardware stop-signal wires through it, per pimm's one-emitter rule), and the hardcoded `TASKS` list dissolves into eval configs under `cfg/eval/real/`. It ships with `privileged={}` until a real ground-truth source (scale) exists, and no success criterion — outcome is the operator's annotation. Shipping a half-wired droid eval in this PR would bypass all of that; design is recorded in the doc.

positronic/task.py
vertix: Timeout landed in 4151a63: a `Task` field (sim factories default 15 s), overridable as `--eval.timeout`; the CLI-level `--simulation_time` is gone. Seed is designed (doc updated): an eval input `--eval.seed`, default `None` = fully random; the scene-owning eval self-seeds `base + n` per trial and self-reports the seed used into episode meta. Implementation lands with step 5 (sim seeding + identity meta). Leaving this open until seed lands.

positronic/policy/harness.py
vertix: Fixed in 4151a63: dropped the archeology comments (the RUN-directive note in harness.py and the entry-point notes in inference.py).

positronic/cfg/eval/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+    policy,
+    trial_count,
+    show_gui,
+    output_dir=None,
+    simulate_inference=False,
+    wrap=default_wrappers,
+):
+    """Run a selected eval (embodiment + task) through the shared inference harness.
+
+    The per-trial time budget is the task's ``timeout`` (override with ``--eval.timeout``).
```
vertix: the override comment does not bring any value and just adds noise

positronic/cfg/eval/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+
+@cfn.config()
+def placeholder():
+    # Lets ``--eval=.sim.positronic.stack_cubes`` resolve relative to this package; never instantiated.
+    raise SystemExit('--eval is required, e.g. --eval=.sim.positronic.stack_cubes')
+
+
+@cfn.config(eval=placeholder, policy=policy_cfg.placeholder, trial_count=1, show_gui=False, wrap=default_wrappers)
+def run(
+    eval: Eval,  # noqa: A002 — the CLI flag is `--eval`, so the parameter must be named `eval`
```
vertix: This noqa is not required anymore, or should be removed. Explain if not possible, but I think it's possible

positronic/cfg/eval/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+    """Run a selected eval (embodiment + task) through the shared inference harness.
+
+    The per-trial time budget is the task's ``timeout`` (override with ``--eval.timeout``).
+    """
+    driver = inference.timed(num_iterations=trial_count, simulation_time=eval.task.timeout, show_gui=show_gui)
+    inference.main(
+        embodiment=eval.embodiment,
+        task=eval.task,
+        policy=policy,
+        driver=driver,
```
vertix, line 23: Wait, haven't we discussed that the start directive will contain both seed and timeout simulation. Did not we discuss that now it should become harness job to watch the time and stop it. Correct me if not, but I think this is the part of it.

positronic/inference.py
vertix, line 115: More open question for discussion, should the inference take Eval as argument? Should the end game be just main(eval, policy) (ok, we also need to think how to deal with wrappers).

Should we pass the eval here right now?

positronic/inference.py
```diff
@@ -120,10 +117,21 @@ def main(
...
+    """Run inference for an embodiment, real or simulated.
+
+    ``task`` (when given) supplies the policy-facing instruction and the privileged
+    ground-truth signals to record; without it the instruction rides the driver.
+    """
     harness = Harness(
-        policy, embodiment, simulate_inference=simulate_inference, on_episode_complete=_completion_sink(policy)
+        policy,
+        embodiment,
+        instruction=task.instruction if task is not None else None,
```
vertix, line 131: I think taks must be not None, how do you imagine it?

positronic/cfg/eval/__init__.py
vertix: Fixed in be3051c: docstring is one line now.

positronic/cfg/eval/__init__.py
vertix: Fixed in be3051c: removed. You're right — `A` (flake8-builtins) isn't in our ruff `select`, so the noqa never did anything.

positronic/cfg/eval/__init__.py
vertix: Half confirm, half correct:

**Timeout — yes, that's the design, scoped to step 4.** Timeout rides the `RUN` directive and the Harness self-terminates at `start + timeout`; the driver becomes a pure trial-sequencer. This PR only makes `task.timeout` the *source of truth*; the enforcement mechanics (the timed driver sleeping) are unchanged here and move into the Harness in step 4 — kept separate because it touches the Harness lifecycle together with the `simulate_inference` → `inference_latency` rename and the typed termination reason.

**Seed — no, deliberately not on `RUN`.** Re-randomization for trial *i* happens at trial *i-1*'s `FINISH`, before trial *i*'s `RUN` context exists, so a directive-borne seed fights the lifecycle ordering (doc, Why #8). Instead the scene-owning eval self-seeds `base + n` on its n-th reset and self-reports the seed used into episode meta (`--eval.seed`, step 5).

positronic/inference.py
vertix: Yes as the end game: once step 6 retires the parallel attended path (EvalUI wrapping the `Eval`), every caller has an `Eval` and this collapses to `main(eval, policy, driver, ...)`. `wrap` stays a separate argument — wrappers are submission-owned (the researcher's pipeline), not eval identity, so they shouldn't fold into the eval.

Not in this PR though: `real`/`phail` have no `Task` yet (the instruction rides the keyboard/EvalUI driver), so passing an `Eval` today means fabricating `Eval(embodiment, task=None)` — that weakens the dataclass contract rather than strengthening the API. The unpacked `(embodiment, task)` is the honest interim.

positronic/inference.py
vertix: Same root as your `main(eval, policy)` question: `task` is `None` exactly on the attended paths (`real`/`phail`), where the driver owns the instruction today (keyboard `--driver.task`, the EvalUI task radio). How I imagine it: step 6 gives those paths real evals (EvalUI wraps the `Eval` and may redefine the instruction per trial via the `RUN` context), at which point `task` becomes required — and `main(eval, policy, ...)` lands in the same move, deleting the `task is None` branches. Forcing it non-optional now would mean a fabricated `Task(instruction=None, timeout=inf)` for the real path, which lies about the contract.

positronic/cfg/eval/__init__.py
vertix: Correction after discussion: you're right on seeds. The sequence is known upfront (`base + i`; with `base=None` a random seed is drawn per trial), and nothing can desync the index — an unattended trial ends only by timeout or the stop-signal, there is no cancel. So the per-trial seed rides `RUN` after all: the eval stays counter-free, and the seed is recorded for free with the rest of the `RUN` context (even fully-random runs become reproducible post-hoc). Doc updated (Why #8 inverted, Principle 12, step 5). Timeout part unchanged: harness-enforced via `RUN` in step 4.

PR #425, opened 2026-06-10
Deliver `timeout` + `inference_latency` on the RUN directive; Harness self-terminates trials

vertix:
## Summary

Step 4 of the eval chain (follow-up to #420): lifecycle knobs and termination become explicit and recorded — and the unattended path loses its driver entirely.

- **The `Harness` holds the `Task`** (replacing its `instruction` arg) and reads the instruction and the per-trial `timeout` straight from it. The timeout never rides `RUN`: the harness sets the `start + task.timeout` deadline **only for trials it starts itself from the plan**, so a directive-driven `RUN` cannot acquire a deadline — the attended surface (keyboard/EvalUI) being its path's termination authority is structural, not a convention.
- **`inference_latency` rides the `RUN` context** and lands in episode meta with the rest of it. `simulate_inference` is renamed to `inference_latency` (`False` = ignore / `True` = measured wall time / `float` = fixed delay, sim-only) and leaves the `Harness` constructor and `inference.main()`; the CLI knob is now `positronic eval run --inference_latency=...`.
- **The Harness self-terminates a self-driven trial** at `start + task.timeout`, stamping `eval.terminated: False` in the episode static meta (time budget exhausted — a future privileged stop-signal will stamp `True`). The deadline is re-checked after the `inference_latency` sleep, so a chunk whose sleep crosses the deadline is dropped, never emitted. Termination is a binary flag; characterizing the end (success/safety/system/...) stays annotation territory — `FINISH(**annotations)` remains free-form, EvalUI/keyboard unchanged.
- **The unattended path has no driver.** A for-loop is not an operator surface: `cfg/eval/run` computes a **trial plan** — one RUN context per trial (`inference_latency`, `eval.trial_index`/`eval.trial_count`; step 5 adds the per-trial seed `base + i` here) — and `Harness(trials=...)` self-drives: whenever idle it starts the next trial and exits once the plan is exhausted. `TimedDriver` is gone (and with it the `simulation.iteration` meta keys).
- **Attended drivers keep their shape**: the anonymous driver tuple becomes a small `Driver` dataclass (`gui`, `directives`, `directive_wrapper`, `control_systems`) for the keyboard/EvalUI surfaces; `main()` takes exactly one of `driver` or `trials`.

## Test plan

- `uv run --locked pytest --no-cov` — full suite passes (551 passed, 7 skipped), including the golden pipeline (RUN-borne `inference_latency` reproduces the exact recorded timings) and the sim integration test, now running **two** planned trials end-to-end with no driver and asserting `eval.terminated`/`eval.trial_index` in episode meta.
- New unit tests: `test_trial_timeout_self_terminates` (self-driven trial ends at `task.timeout` → ds STOP carries `eval.terminated: False`, devices homed), `test_trial_plan_self_drives` (driverless harness runs the plan: one episode per entry, indices and the task instruction recorded), `test_timeout_crossed_during_latency_sleep_drops_chunk` (post-deadline chunk never reaches the drivers).
- Smoke: `positronic eval run --help`, `positronic-inference sim --help`, `utilities/check_packaging.py` (60 modules OK).

### DIFF
--- a/positronic/cfg/eval/__init__.py
+++ b/positronic/cfg/eval/__init__.py
@@ -13,15 +13,11 @@ def placeholder():
 
 
 @cfn.config(eval=placeholder, policy=policy_cfg.placeholder, trial_count=1, show_gui=False, wrap=default_wrappers)
-def run(eval: Eval, policy, trial_count, show_gui, output_dir=None, simulate_inference=False, wrap=default_wrappers):
+def run(eval: Eval, policy, trial_count, show_gui, output_dir=None, inference_latency=False, wrap=default_wrappers):
     """Run a selected eval (embodiment + task) through the shared inference harness."""
-    driver = inference.timed(num_iterations=trial_count, simulation_time=eval.task.timeout, show_gui=show_gui)
+    driver = inference.sequencer(
+        trial_count=trial_count, timeout=eval.task.timeout, inference_latency=inference_latency, show_gui=show_gui
+    )
     inference.main(
-        embodiment=eval.embodiment,
-        task=eval.task,
-        policy=policy,
-        driver=driver,
-        output_dir=output_dir,
-        simulate_inference=simulate_inference,
-        wrap=wrap,
+        embodiment=eval.embodiment, task=eval.task, policy=policy, driver=driver, output_dir=output_dir, wrap=wrap
     )
--- a/positronic/cfg/eval/__init__.py
+++ b/positronic/cfg/eval/__init__.py
@@ -15,9 +15,23 @@ def placeholder():
 @cfn.config(eval=placeholder, policy=policy_cfg.placeholder, trial_count=1, show_gui=False, wrap=default_wrappers)
 def run(eval: Eval, policy, trial_count, show_gui, output_dir=None, inference_latency=False, wrap=default_wrappers):
     """Run a selected eval (embodiment + task) through the shared inference harness."""
-    driver = inference.sequencer(
-        trial_count=trial_count, timeout=eval.task.timeout, inference_latency=inference_latency, show_gui=show_gui
-    )
+    # The trial plan: one RUN context per trial, consumed by the self-driving Harness.
+    # (Step 5 adds the per-trial seed `base + i` here.)
+    trials = [
+        {
+            'timeout': eval.task.timeout,
+            'inference_latency': inference_latency,
+            'eval.trial_index': i,
+            'eval.trial_count': trial_count,
+        }
+        for i in range(trial_count)
+    ]
     inference.main(
-        embodiment=eval.embodiment, task=eval.task, policy=policy, driver=driver, output_dir=output_dir, wrap=wrap
+        embodiment=eval.embodiment,
+        task=eval.task,
+        policy=policy,
+        trials=trials,
+        show_gui=show_gui,
+        output_dir=output_dir,
+        wrap=wrap,
     )
--- a/positronic/cfg/eval/__init__.py
+++ b/positronic/cfg/eval/__init__.py
@@ -16,14 +16,8 @@ def placeholder():
 def run(eval: Eval, policy, trial_count, show_gui, output_dir=None, inference_latency=False, wrap=default_wrappers):
     """Run a selected eval (embodiment + task) through the shared inference harness."""
     # The trial plan: one RUN context per trial, consumed by the self-driving Harness.
-    # (Step 5 adds the per-trial seed `base + i` here.)
     trials = [
-        {
-            'timeout': eval.task.timeout,
-            'inference_latency': inference_latency,
-            'eval.trial_index': i,
-            'eval.trial_count': trial_count,
-        }
+        {'inference_latency': inference_latency, 'eval.trial_index': i, 'eval.trial_count': trial_count}
         for i in range(trial_count)
     ]
     inference.main(
--- a/positronic/policy/harness.py
+++ b/positronic/policy/harness.py
@@ -163,29 +163,26 @@ def default_wrappers(clock: pimm.Clock) -> PolicyWrapper:
 class Harness(pimm.ControlSystem):
     """Control system that manages episode lifecycle and forwards trajectories to drivers.
 
-    The harness handles directives (RUN/STOP/FINISH/HOME) and dataset recording.
-    All inference intelligence (scheduling, error recovery, blending, absolute
-    time stamping) lives in the policy/session layer — the harness just calls
-    the session, demuxes the action dicts into per-channel trajectories, and
-    emits.
+    The harness handles directives (RUN/STOP/FINISH/HOME) and dataset recording. All inference
+    intelligence (scheduling, error recovery, blending, absolute time stamping) lives in the
+    policy/session layer — the harness just calls the session, demuxes the action dicts into
+    per-channel trajectories, and emits.
 
-    ``RUN`` may carry ``inference_latency`` (sim-only inference-cost simulation)
-    in its context. A ``trials`` plan (a sequence of RUN contexts) makes the
-    harness self-driving: whenever it is idle it starts the next trial itself —
-    bounded by the task's ``timeout`` — and exits once the plan is exhausted, so
-    the unattended path needs no driver at all. Attended drivers own episode
+    ``RUN`` may carry ``inference_latency`` (sim-only inference-cost simulation) in its context.
+    A ``trials`` plan (a sequence of RUN contexts) makes the harness self-driving: whenever it is
+    idle it starts the next trial itself — bounded by the task's ``timeout`` — and exits once the
+    plan is exhausted, so the unattended path needs no driver at all. Attended drivers own episode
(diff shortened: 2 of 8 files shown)


positronic/policy/harness.py
```diff
@@ -422,6 +435,12 @@ def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[p
             if directive_msg.updated:
                 running, recording = yield from self._handle_directive(directive_msg.data, clock, recording)
 
+            if running and self._deadline is not None and clock.now() >= self._deadline:
+                # Trial time budget exhausted: self-terminate. ``eval.terminated`` is False —
+                # the trial ran out of time rather than being terminated by a stop-signal.
+                finish = Directive.FINISH(**{'eval.terminated': False})
+                running, recording = yield from self._handle_directive(finish, clock, recording)
```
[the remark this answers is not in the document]
vertix: Fixed in bdb4a58: `_step` now rechecks the deadline after the latency sleep (covers a slow wall-clock inference call too) and drops the chunk instead of emitting it — the run loop then fires the timeout FINISH on its next cycle. Added a regression test (`test_timeout_crossed_during_latency_sleep_drops_chunk`) that fails without the recheck.

positronic/policy/harness.py
```diff
@@ -392,17 +405,24 @@ def _step(self, clock: pimm.Clock) -> Generator[pimm.Sleep, None, None]:
         actions = self._session(frozen_view(obs))
         if actions is None:
             return
-        if self.simulate_inference is True:  # bool is an int subclass — check identity first
+        if self._inference_latency is True:  # bool is an int subclass — check identity first
```
vertix: Can we please have a small method that returns `delay`, so that we do `delay = self.???` then `if delay> 0:`

positronic/cfg/eval/__init__.py
```diff
@@ -13,15 +13,11 @@ def placeholder():
 
 
 @cfn.config(eval=placeholder, policy=policy_cfg.placeholder, trial_count=1, show_gui=False, wrap=default_wrappers)
-def run(eval: Eval, policy, trial_count, show_gui, output_dir=None, simulate_inference=False, wrap=default_wrappers):
+def run(eval: Eval, policy, trial_count, show_gui, output_dir=None, inference_latency=False, wrap=default_wrappers):
     """Run a selected eval (embodiment + task) through the shared inference harness."""
-    driver = inference.timed(num_iterations=trial_count, simulation_time=eval.task.timeout, show_gui=show_gui)
+    driver = inference.sequencer(
+        trial_count=trial_count, timeout=eval.task.timeout, inference_latency=inference_latency, show_gui=show_gui
```
vertix: There's clear duplication. Eval carries over the timeout now, so it seems that harness can read it directly from Eval, and driver does not need to know the timeout, it just becomes part of Eval (which I think makes sense)

positronic/cfg/eval/__init__.py
```diff
@@ -13,15 +13,25 @@ def placeholder():
 
 
 @cfn.config(eval=placeholder, policy=policy_cfg.placeholder, trial_count=1, show_gui=False, wrap=default_wrappers)
-def run(eval: Eval, policy, trial_count, show_gui, output_dir=None, simulate_inference=False, wrap=default_wrappers):
+def run(eval: Eval, policy, trial_count, show_gui, output_dir=None, inference_latency=False, wrap=default_wrappers):
     """Run a selected eval (embodiment + task) through the shared inference harness."""
-    driver = inference.timed(num_iterations=trial_count, simulation_time=eval.task.timeout, show_gui=show_gui)
+    # The trial plan: one RUN context per trial, consumed by the self-driving Harness.
+    # (Step 5 adds the per-trial seed `base + i` here.)
```
vertix: Don't leave this kind of comments for future or the past. For the future only todos, and never for the past

positronic/policy/harness.py
vertix: Fixed in 0526ea3: extracted `_inference_delay(wall_start)`; `_step` now reads `delay = self._inference_delay(wall_start)` / `if delay > 0.0:`.

positronic/cfg/eval/__init__.py
vertix: Fixed in 0526ea3: the `Harness` now takes the `Task` itself (replacing the `instruction` arg) and reads both the instruction and the timeout from it. `timeout` no longer rides `RUN` or the trial plan at all — the harness sets the `start + task.timeout` deadline only for trials it starts itself from the plan. Nice side effect: an attended (directive-driven) `RUN` can't acquire a deadline, so the UI being its path's termination authority is now structural rather than a convention.

positronic/cfg/eval/__init__.py
vertix: Fixed in 0526ea3: dropped the comment.

positronic/cfg/eval/__init__.py
```diff
@@ -13,15 +13,19 @@ def placeholder():
 
 
 @cfn.config(eval=placeholder, policy=policy_cfg.placeholder, trial_count=1, show_gui=False, wrap=default_wrappers)
-def run(eval: Eval, policy, trial_count, show_gui, output_dir=None, simulate_inference=False, wrap=default_wrappers):
+def run(eval: Eval, policy, trial_count, show_gui, output_dir=None, inference_latency=False, wrap=default_wrappers):
     """Run a selected eval (embodiment + task) through the shared inference harness."""
-    driver = inference.timed(num_iterations=trial_count, simulation_time=eval.task.timeout, show_gui=show_gui)
+    # The trial plan: one RUN context per trial, consumed by the self-driving Harness.
+    trials = [
+        {'inference_latency': inference_latency, 'eval.trial_index': i, 'eval.trial_count': trial_count}
```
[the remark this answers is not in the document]
vertix: Declining — this is deliberate in 0526ea3.

The timeout is a harness lifecycle bound, not policy input or per-trial data, so it doesn't belong in the trial context:

1. Context keys flow into policy inputs (`Harness._build_obs` does `inputs.update(self.context)`). The policy has no use for the time budget, and a `timeout` key riding the RUN context would look like a behavioral knob while being inert — the deadline comes from `task.timeout` held by the `Harness`, and attended (directive-driven) RUNs intentionally get no deadline at all. That ambiguity is exactly what this commit removed.
2. Reproducibility is covered at the right level: `inference.main` writes `run_metadata_*.yaml` (full `sys.argv` including any `--eval.task.timeout` override, plus git state/diff), which is the documented record for reconstructing runs. The timeout is run-level data anyway — one `Task` per run, shared by every trial — so duplicating it into each episode's static meta adds nothing. Per episode, timed-out trials already record `eval.terminated: False`, distinguishing timeout from stop-signal termination.

positronic/policy/harness.py
```diff
@@ -381,28 +405,28 @@ def _step(self, clock: pimm.Clock) -> Generator[pimm.Sleep, None, None]:
...
-            delay = float(self.simulate_inference)
-        else:
-            delay = 0.0
+        delay = self._inference_delay(wall_start)
         if delay > 0.0:
             yield pimm.Sleep(delay)
             actions = [{**a, 'timestamp': a['timestamp'] + delay} for a in actions]
             self._bump_schedule_end(delay)
 
+        # Recheck the deadline: the latency sleep (or a slow inference call on a
```
vertix: Pls use cheaper model to recact all the comments you made in this PR, so that their width matches the width allowed in our repo

positronic/policy/harness.py
```diff
@@ -421,6 +445,19 @@ def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[p
             directive_msg = self.directive.read()
             if directive_msg.updated:
                 running, recording = yield from self._handle_directive(directive_msg.data, clock, recording)
+            elif not running and self._trials is not None:
```
vertix: The main loop became very verbose now, is it smth we have to take, or can we polish it so it's less wordy?

positronic/policy/harness.py
vertix: Fixed in a5b7b50 — re-wrapped (via a cheaper model) all comments and docstrings this PR added to fill the repo's 120-column width: the `Harness` class docstring and `__init__` state comments, `_bump_schedule_end`/`main()` docstrings, the `_step` latency + deadline-recheck comments, and the two test comments. Wording unchanged, only line breaks.

positronic/policy/harness.py
vertix: Fixed in a5b7b50. The verbosity was the `running, recording` tuple threaded through every `_handle_directive` call. Moved both to instance state (`self._running`/`self._recording`) next to the rest of the lifecycle state (`_session`, `_deadline`, `context`): `_handle_directive` now updates them directly, so all three call sites collapse to a plain `yield from self._handle_directive(...)`, and the timeout branch lost its temp variable. The loop body is now one screen of mostly single-line statements.

PR #432, opened 2026-06-14
DreamZero: AR video context at serving + full-finetune warm-start

vertix:
## Summary

Makes DreamZero serve correctly and lets a full finetune continue training, plus the Nebius wiring to run both.

- **`TemporalFrameStack` at serving** (`positronic/policy/harness.py`) — feeds DreamZero its autoregressive multi-frame video context (4 frames at the trained `[-23, -16, -8, 0]` cadence) at serving time. The harness otherwise only forwards an observation to the policy at re-query boundaries, so the server saw the wrong temporal context. `dreamzero_wrappers` composes it with the scheduling wrappers, and the codec's `_encode_image` now resizes a per-frame stack.
- **Full-finetune warm-start** (`positronic/vendors/dreamzero/train.py`) — `init_from_checkpoint` loads a prior run's weights into the freshly-built model (fresh optimizer + schedule). `save_only_model=true` (needed to keep h100x1 checkpoints small) precludes a true DeepSpeed resume, so warm-start is the way to train an existing checkpoint longer. Also exposes `lora_rank` / `lora_alpha` / `lora_adapt_vision` knobs.
- **Nebius wiring** (`workflows/nebius/{serve,train}.sh`) — adds the `dreamzero` vendor case and makes the job timeout overridable via `NEBIUS_JOB_TIMEOUT`.
- **S3 doc** (`docs/nebius-s3-checkpoint-failures.md`) — records the intermittent Nebius S3 checkpoint-upload failures seen during training (partial uploads that look complete; how to detect and mitigate).

## Test plan

- `ruff` clean; 52 unit tests pass (`policy/tests/test_wrappers.py`, `test_harness.py`, `vendors/dreamzero/tests/test_codecs.py`).
- Frame-pump validated closed-loop in sim on a full-finetune checkpoint: drift roughly halved (~1.60 → ~0.76 rad) and per-step command jitter dropped to ~0.012 vs the single-frame serving path.
- Warm-start: a full-finetune continuation from `checkpoint-9000` is running on Nebius; end-to-end completion not yet confirmed.

## Notes for review

- The `lora_adapt_vision` vision-LoRA path (and its Dockerfile injection) and the norm-monitoring Dockerfile patch are exploratory; LoRA paths collapsed in testing and full finetune is the viable model. Happy to drop either before merge.


positronic/policy/harness.py
```diff
@@ -155,6 +157,63 @@ def wrap_session(self, inner: Session, context):
...
+
+        def __call__(self, obs):
+            now = self._clock.now()
+            stacked = dict(obs)
+            for key in self._image_keys:
+                frame = obs.get(key)
+                if isinstance(frame, np.ndarray) and frame.ndim == 3:
+                    buf = self._buffers[key]
+                    if not buf or now - buf[-1][0] >= self._min_dt_sec:
+                        buf.append((now, np.asarray(frame)))
```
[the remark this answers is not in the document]
vertix: Fixed in da2f129.

`np.asarray(frame)` returns the input ndarray unchanged when it's already an array, so every buffered entry aliased the same reused shared-memory camera buffer (`NumpySMAdapter.array`, returned by `Serializers.camera_images`). Since the producer renders the next frame into that buffer in place each tick, by `np.stack` time all slots pointed at the latest frame — the model received N copies of "now" instead of the `[-23, -16, -8, 0]` offsets.

Switched to `np.array(frame)` (forces a copy) so each buffered entry is an immutable snapshot, with a comment noting the shared-memory reuse. Confirmed against the upstream DreamZero reference client (`serve_dreamzero_wan22.py` + `test_client_AR.py`), which sends 4 distinct frames per call at those offsets — so the stack must hold distinct historical frames. Re-ran the sim eval through the corrected pump; trajectories changed, verifying the conditioning now propagates.

positronic/policy/harness.py
```diff
@@ -155,6 +157,66 @@ def wrap_session(self, inner: Session, context):
...
+            stacked = dict(obs)
+            for key in self._image_keys:
+                frame = obs.get(key)
+                if isinstance(frame, np.ndarray) and frame.ndim == 3:
+                    buf = self._buffers[key]
+                    if not buf or now - buf[-1][0] >= self._min_dt_sec:
+                        # Copy: camera frames are views into a shared-memory buffer the producer reuses
+                        # each tick, so storing the view would alias every slot to the latest frame and
+                        # the stack would be N copies of "now" instead of the offset history.
+                        buf.append((now, np.array(frame)))
```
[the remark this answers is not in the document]
vertix: Fixed in 54da710.

After appending, the buffer now drops everything older than the oldest sampled offset (`now + min(offsets_sec)`), keeping a single frame just beyond it so nearest-neighbor sampling still brackets that offset. Bounds the buffer to the ~1.6s window the sampler actually reads instead of letting it grow for the whole episode.

positronic/vendors/dreamzero/train.py
```diff
@@ -149,6 +158,18 @@ def main(
     ]
     if bb['frame_seqlen'] is not None:
         command.append(f'frame_seqlen={bb["frame_seqlen"]}')
+    if lora_rank is not None:
+        command.append(f'action_head_cfg.config.lora_rank={lora_rank}')
+    if lora_alpha is not None:
+        command.append(f'action_head_cfg.config.lora_alpha={lora_alpha}')
+    if lora_adapt_vision:
+        # `+` appends the key: it lives on the patched dataclass, not in the action-head YAML struct.
+        command.append('+action_head_cfg.config.lora_adapt_vision=true')
```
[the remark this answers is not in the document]
vertix: Fixed in 152f9a6 by removing the vision-LoRA path entirely rather than guarding it.

`lora_rank` / `lora_alpha` / `lora_adapt_vision` came from an exploration that didn't pan out and aren't used by any preset (the warm-start run is a full finetune). Rather than carry a wan22-incompatible knob into the merge, the params and the `Dockerfile.dreamzero` visual-tower patch are dropped. LoRA rank/alpha remain settable via `extra_args` Hydra overrides if needed later. The net PR diff no longer contains the flag, so the wan22 footgun is gone.

positronic/policy/harness.py
```diff
@@ -155,6 +157,71 @@ def wrap_session(self, inner: Session, context):
...
+                        # Drop frames older than the oldest sampled offset (keep one just beyond it so
+                        # nearest-neighbor sampling still brackets it): the sampler never looks further
+                        # back, so without this the buffer grows for the whole episode.
+                        cut = next(i for i, (t, _) in enumerate(buf) if t >= now + min(self._offsets_sec))
+                        del buf[: max(cut - 1, 0)]
+                    times = np.array([t for t, _ in buf])
+                    idxs = [int(np.argmin(np.abs(times - (now + off)))) for off in self._offsets_sec]
+                    stacked[key] = np.stack([buf[i][1] for i in idxs])
```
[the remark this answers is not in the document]
vertix: Fixed in efe1c49.

`RemoteSession._prepare_obs` now resizes 4-D temporal stacks the same way it already resized single frames — each frame is fit to the server's advertised size (`image_sizes`) before sending — so a `(T, H, W, 3)` stack of hd720 cameras ships downscaled instead of full-resolution. Kept the resize in `RemoteSession` (it owns the server's expected sizes) rather than duplicating it into the wrapper.

positronic/policy/remote.py
```diff
@@ -31,17 +31,26 @@ def _resize_to(image: np.ndarray, width: int, height: int) -> np.ndarray:
...
+        scale = min(1.0, tw / w, th / h)
+        return RemoteSession._resize_to(image, int(w * scale), int(h * scale))
+
     def _prepare_obs(self, obs: dict[str, Any]) -> dict[str, Any]:
         result = {}
         for key, value in obs.items():
-            if isinstance(value, np.ndarray) and value.ndim == 3 and value.shape[2] == 3:
+            # Resize single RGB frames and temporal stacks of them alike (TemporalFrameStack emits a
+            # (T, H, W, 3) stack), so a stack of hd720 frames isn't shipped full-resolution.
+            if isinstance(value, np.ndarray) and value.shape[-1] == 3 and value.ndim in (3, 4):
```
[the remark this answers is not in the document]
vertix: Fixed in 6767d21. Good catch — this was a regression from the previous commit.

Reordered the guard to `value.ndim in (3, 4) and value.shape[-1] == 3` so the rank is checked first; `and` short-circuits, so `shape[-1]` is only read once `ndim` is known to be 3 or 4. Scalar (0-D) and other non-image features now pass through untouched again.

positronic/policy/harness.py
```diff
@@ -155,6 +157,71 @@ def wrap_session(self, inner: Session, context):
...
+            self._offsets_sec = offsets_sec
+            self._min_dt_sec = min_dt_sec
+            self._buffers: dict[str, list[tuple[float, np.ndarray]]] = {k: [] for k in image_keys}
+
+        def __call__(self, obs):
+            now = self._clock.now()
+            stacked = dict(obs)
+            for key in self._image_keys:
+                frame = obs.get(key)
+                if isinstance(frame, np.ndarray) and frame.ndim == 3:
```
vertix: The keys MUST be images. We should fail loudly if they are not. Again, you apply defensive programming where you should not

positronic/policy/harness.py
```diff
@@ -155,6 +157,71 @@ def wrap_session(self, inner: Session, context):
...
+                    buf = self._buffers[key]
+                    if not buf or now - buf[-1][0] >= self._min_dt_sec:
```
vertix: How could it be that buf is None? Isn't it defensive again?

positronic/policy/harness.py
```diff
@@ -155,6 +157,71 @@ def wrap_session(self, inner: Session, context):
         return ErrorRecovery._Session(inner, self._clock)
 
 
+class TemporalFrameStack(PolicyWrapper):
```
vertix: This PR showed that we need to make a refactor for wrappers code
* extract all wrappers to wrappers.py
* All wrappers should accept Clock, but clock must be not a Clock instance (don't want pimm to get into policy), but just a callable that returns clock (i.e. now()).
* We should also define configs for the wrappers. They take the arguments (like imаge keys here), and return the clock-parametrised factories.
* Also, pls extract the buffers logic into a separate utility method or class, so that we don't overcomplicate this wrapper with it
* Consider to simplify sessions with carrying the link to wrapper, so that we don't have to pass all the arguments around

positronic/vendors/dreamzero/codecs.py
```diff
@@ -238,3 +247,15 @@ def _ik_dreamzero_action(solver: str):
...
+
+def dreamzero_wrappers(clock: pimm.Clock) -> PolicyWrapper:
+    """Eval wrapper pipeline for DreamZero: buffer frames and feed the server the AR video context.
+
+    Drop-in for the eval ``wrap`` param. Stacks the wrist/exterior cameras at
+    ``FRAME_STACK_OFFSETS_SEC`` outside the scheduling wrappers, so the server receives multi-frame
+    context at the trained cadence. Pair with full-chunk execution (codec ``horizon`` covering the
+    whole 24-step chunk) so the executed window matches the frame offsets.
+    """
+    frame_stack = TemporalFrameStack(clock, ('image.wrist', 'image.exterior'), FRAME_STACK_OFFSETS_SEC)
```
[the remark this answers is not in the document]
vertix: Fixed in 5b8630d. The DreamZero composed codecs (`joints`, `joints_ik`) now set `horizon=None`, so the full 24-action chunk executes instead of being truncated to 15 by the compose default of 1.0s. Re-query now aligns with the frame-stack window, with no per-run override needed.

positronic/policy/harness.py
vertix: Fixed in 5b8630d. Dropped the `isinstance(frame, np.ndarray) and frame.ndim == 3` guard — the wrapper now reads `obs[key]` directly and trusts that `image_keys` map to image observations, failing loudly (KeyError) if one is missing or misconfigured. The logic moved to `TemporalFrameStack._Session` / `_FrameBuffer` in the new `positronic/policy/wrappers.py`.

positronic/policy/harness.py
vertix: Addressed in 5b8630d. `buf` is `self._buffers[key]` — always a list, never None; `not buf` is the empty-buffer (first-frame) case guarding `buf[-1]`, not a defensive None-check. It now lives inside `_FrameBuffer.append` (the extracted buffer utility), so the wrapper body no longer carries it.

positronic/policy/harness.py
vertix: Done in 5b8630d. The wrappers now live in `positronic/policy/wrappers.py`:

- All three take a `now` callable (`Callable[[], float]`), not a `pimm.Clock`, so pimm stays out of `policy/`. The harness passes `clock.now`.
- `@cfn.config` factories per wrapper (`chunked_schedule`, `error_recovery`, `temporal_frame_stack`) returning `now -> wrapper` factories, plus a `compose` helper; `dreamzero_wrappers` composes them.
- The frame-buffer logic is extracted into a `_FrameBuffer` utility, so `TemporalFrameStack` stays thin.
- Sessions carry a back-reference to their wrapper instead of threading every arg through `_Session.__init__`.

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,187 @@
+"""Composable policy wrappers — scheduling, error recovery, and temporal frame stacking.
+
+Wrappers are clock-aware concerns layered around a policy at serving time. They are clock-free
```
vertix: This doc says too much about clock (and clock-free).
The clock appear only in `wrap`, so the amount of attention to it should be decreased accordingly

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,187 @@
+"""Composable policy wrappers — scheduling, error recovery, and temporal frame stacking.
```
vertix, line 1: We speak too much about particular implementations that we have here. Should we rather import the definition of classes here as well, and also have particular implementations?

Only if it's possible to decouple

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,187 @@
...
+
+Wrappers are clock-aware concerns layered around a policy at serving time. They are clock-free
+recipes composed with ``|`` (left is outermost), exactly like codecs; the harness supplies the
+runtime clock when it applies the pipeline (``wrap(policy, now)``), and only the per-session
+``_Session`` holds ``now`` — a ``Callable[[], float]`` returning the current time in seconds.
+Each wrapper exposes a ``@cfn.config`` so pipelines can be selected and overridden from the CLI.
+"""
+
+import configuronic as cfn
```
vertix: All configurations should go to cfg module. No code in policy should depend on configs. If it is no the case in the base PR, let me know

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,187 @@
...
+
+    class _Session(DelegatingSession):
+        def __init__(self, inner: Session, wrapper: 'TemporalFrameStack', now: Now):
+            super().__init__(inner)
+            self._wrapper = wrapper
+            self._now = now
+            self._buffer = _FrameBuffer(wrapper._image_keys, wrapper._offsets_sec, wrapper._min_dt_sec)
+
+        def __call__(self, obs):
+            now = self._now()
```
vertix: Haven't we discussed that only ChunkedSchedule really need now, and that for this one and for ErrorRecovery should use the one from observation, and that it is even better to do so?

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,187 @@
...
+            stacked = dict(obs)
+            for key in self._wrapper._image_keys:
+                self._buffer.append(key, now, obs[key])
```
vertix: We always push the dict, and stack the dict? Why do we have this loop here, when everything can be hidden in buffer, and we don't even actually need the list of keys, we can just use all of them

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,187 @@
...
+        self._buffers = {k: [] for k in self._buffers}
+
+    def append(self, key: str, now: float, frame: np.ndarray):
+        buf = self._buffers[key]
+        if not buf or now - buf[-1][0] >= self._min_dt_sec:
+            # Copy: camera frames are views into a shared-memory buffer the producer reuses each tick,
+            # so storing the view would alias every slot to the latest frame.
+            buf.append((now, np.array(frame)))
+            cut = next(i for i, (t, _) in enumerate(buf) if t >= now + min(self._offsets_sec))
+            del buf[: max(cut - 1, 0)]
```
vertix: If we delete from the head, should it be deque rather than a list?

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,187 @@
...
+    def reset(self):
+        self._buffers = {k: [] for k in self._buffers}
+
+    def append(self, key: str, now: float, frame: np.ndarray):
+        buf = self._buffers[key]
+        if not buf or now - buf[-1][0] >= self._min_dt_sec:
+            # Copy: camera frames are views into a shared-memory buffer the producer reuses each tick,
+            # so storing the view would alias every slot to the latest frame.
+            buf.append((now, np.array(frame)))
+            cut = next(i for i, (t, _) in enumerate(buf) if t >= now + min(self._offsets_sec))
```
vertix: should not we accept only ordered offeset_sec list?
Overall, we should rely (and verify) that observations come ordered in time, then we could use sorted arrays mechanics, similar to what we use in dataset/signals

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,187 @@
...
+            for key in self._wrapper._image_keys:
+                self._buffer.append(key, now, obs[key])
+                stacked[key] = self._buffer.sample(key, now)
+            return self._inner(stacked)
+
+        def cancel(self):
+            self._buffer.reset()
+            super().cancel()
+
+    def __init__(self, image_keys: tuple[str, ...], offsets_sec: tuple[float, ...], min_dt_sec: float = 0.18):
```
vertix: Why do we need min_dt_sec at all?

positronic/policy/recording.py
```diff
@@ -332,7 +332,7 @@ def __init__(self, rec: Recorder, name: str):
         self._rec = rec
         self._name = name
 
-    def wrap_session(self, inner: Session, context) -> Session:
+    def wrap_session(self, inner: Session, context, now) -> Session:
```
vertix, line 341: Should not we use this `now` in _RecordingTapSession? Does it give us extra information?

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,187 @@
...
+            self._wrapper = wrapper
+            self._now = now
+            self._buffer = _FrameBuffer(wrapper._image_keys, wrapper._offsets_sec, wrapper._min_dt_sec)
+
+        def __call__(self, obs):
+            now = self._now()
+            stacked = dict(obs)
+            for key in self._wrapper._image_keys:
+                self._buffer.append(key, now, obs[key])
+                stacked[key] = self._buffer.sample(key, now)
```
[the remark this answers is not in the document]
vertix: The buffer records cheaply every tick (a dict of copies), but the stack is still materialized each tick. Deferring it until a re-query is due would require this wrapper to know `ChunkedSchedule`'s gate state *before* calling inner — yet it must substitute the obs before that call, and it deliberately sits outside the scheduler so it sees every frame. So a clean deferral conflicts with the composition. Leaving this open: if idle-tick `np.stack` cost shows up in practice, the cleaner fix is likely a lazy stacked view rather than inverting the wrapper order.

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,187 @@
...
+        self._offsets_sec = offsets_sec
+        self._min_dt_sec = min_dt_sec
+        self._buffers: dict[str, list[tuple[float, np.ndarray]]] = {k: [] for k in image_keys}
+
+    def reset(self):
+        self._buffers = {k: [] for k in self._buffers}
+
+    def append(self, key: str, now: float, frame: np.ndarray):
+        buf = self._buffers[key]
+        if not buf or now - buf[-1][0] >= self._min_dt_sec:
```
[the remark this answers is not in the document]
vertix: Fixed in c3b4f72: dropped `min_dt_sec` entirely. The buffer now stores every tick (pruned to the sampled window), so the offset-`0` frame is always the current observation — the up-to-180ms skew is gone.

positronic/policy/wrappers.py
vertix: Fixed in c3b4f72: trimmed the module docstring. It now states only that `ChunkedSchedule` needs the live clock (others read the observation's `obs_time_ns`), instead of dwelling on clock-free composition.

positronic/policy/wrappers.py
vertix: The base definitions already live in `base.py` (`PolicyWrapper`, `Session`, `DelegatingSession`); `wrappers.py` holds only the concrete serving wrappers, and the configs now live in `cfg/wrappers.py` (c3b4f72). I also trimmed the implementation-heavy docstring. If you had a further split in mind (e.g. an interface module separate from implementations), let me know — leaving open.

positronic/policy/wrappers.py
vertix: Fixed in c3b4f72: moved the wrapper `@cfn.config` factories to a new `cfg/wrappers.py`. `positronic/policy/wrappers.py` no longer imports configuronic, and a grep confirms `positronic/policy/` no longer depends on configs at all — the only such dependency was this newly-extracted file.

positronic/vendors/dreamzero/codecs.py
```diff
@@ -236,5 +245,24 @@ def _ik_dreamzero_action(solver: str):
...
+
+_frame_stack = wrappers.temporal_frame_stack.override(
+    image_keys=('image.wrist', 'image.exterior'), offsets_sec=FRAME_STACK_OFFSETS_SEC
+)
+
+
+@cfn.config(
+    frame_stack=_frame_stack, error_recovery=wrappers.error_recovery, chunked_schedule=wrappers.chunked_schedule
+)
+def dreamzero_wrappers(frame_stack, error_recovery, chunked_schedule):
```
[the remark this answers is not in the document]
vertix: Addressed in c3b4f72 by documenting the override: the DreamZero README Quick Start now passes `--policy.codec=@positronic.vendors.dreamzero.codecs.joints` and `--wrap=@positronic.vendors.dreamzero.codecs.dreamzero_wrappers`, so the documented path selects the AR frame stack. We intentionally keep wrapper selection on the CLI — the wrapper set is a per-run serving choice (e.g. `--wrap.chunked_schedule=...` to swap in RTC) rather than a baked preset.

positronic/vendors/dreamzero/codecs.py
```diff
@@ -214,9 +221,11 @@ def dreamzero_action(tgt_joints_key: str, tgt_grip_key: str, num_joints: int):
...
-# Composed codec: DreamZero observation + action + timing
+# Composed codec: DreamZero observation + action + timing.
+# horizon=None executes the full 24-action chunk DreamZero returns, aligning the re-query with the
+# frame-stack window; a finite horizon (the compose default is 1.0s) would truncate the chunk.
 _action = dreamzero_action.override(tgt_joints_key='robot_command.joints', tgt_grip_key='target_grip')
-joints = codecs.compose.override(obs=dreamzero_obs, action=_action, fps=15.0)
+joints = codecs.compose.override(obs=dreamzero_obs, action=_action, fps=15.0, horizon=None)
 
 _traj_action = dreamzero_action.override(tgt_joints_key='robot_state.q', tgt_grip_key='grip')
 joints_traj = codecs.compose.override(obs=dreamzero_obs, action=_traj_action, fps=15.0)
```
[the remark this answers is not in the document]
vertix: Fixed in c3b4f72: `joints_traj` now sets `horizon=None`, matching `joints`/`joints_ik`, so the full 24-step chunk executes and stays aligned with the frame-stack window.

positronic/policy/wrappers.py
vertix: Fixed in c3b4f72, exactly as we discussed: `ErrorRecovery` and `TemporalFrameStack` now read `obs['obs_time_ns']`; only `ChunkedSchedule` takes the injected live clock, because it alone must anchor the chunk to inference *completion*, which the pre-inference observation stamp cannot provide.

positronic/policy/wrappers.py
vertix: Fixed in c3b4f72: the per-key loop moved into `_FrameBuffer` — the session pushes the image dict and gets the stacked dict back. Kept `image_keys` explicit (per your steer): we stack only the cameras, not state vectors like `joint_position`/`grip`.

positronic/policy/wrappers.py
vertix: Fixed in c3b4f72: `_FrameBuffer` uses a `deque` with O(1) `popleft` from the head.

positronic/policy/wrappers.py
vertix: Fixed in c3b4f72: the buffer is time-ordered and sampling uses `np.searchsorted` to pick the nearest frame per offset; `offsets_sec` is taken ascending. This mirrors the sorted-array approach in `dataset/signals`.

positronic/policy/wrappers.py
vertix: Fixed in c3b4f72: removed `min_dt_sec`. It throttled storage but at 0.18s it was coarser than the 15fps offset spacing, so adjacent offsets could alias to the same stored frame. Storing every tick (pruned to the sampled window) samples exactly and also removes the offset-`0` skew.

positronic/policy/recording.py
vertix: No extra information: the tap stamps its rerun timelines from the observation's own `wall_time_ns`/`obs_time_ns` (when the obs was assembled) — `now()` would just re-read the same clock a few microseconds later. With the obs-time change, `now` is only consumed by `ChunkedSchedule`; the tap ignores it. Leaving open in case you'd want the tap to record an additional "tap-entered" wall time.

positronic/policy/wrappers.py
vertix: I know they live in base.py, my question was whether we should move them here?

positronic/vendors/dreamzero/README.md
```diff
@@ -28,10 +28,16 @@ uv run positronic-inference sim \
   --policy=.remote \
   --policy.host=vm-dreamzero \
   --policy.port=8000 \
+  --policy.codec=@positronic.vendors.dreamzero.codecs.joints \
```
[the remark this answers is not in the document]
vertix: Fixed in 5adce3e: dropped `--policy.codec=…joints` from the quick-start. You're right — `server` defaults to `codec=joints`, so the server already does `codec.wrap(base_policy)` (encodes the obs, decodes the action); adding the codec client-side double-encodes and the first request fails on the raw-key read. The client now sends raw observations and receives decoded joint commands; only `--wrap` stays client-side, since `TemporalFrameStack` must run every control tick to record frames. Added a README note spelling this out.

workflows/nebius/train.sh
```diff
@@ -54,8 +56,9 @@ case "$VENDOR" in
   lerobot)       IMAGE="positro/positronic:${IMAGE_TAG}"; EXTRA="--extra lerobot " ;;
   openpi)        IMAGE="positro/openpi:${IMAGE_TAG}";     EXTRA="" ;;
   gr00t)         IMAGE="positro/gr00t:${IMAGE_TAG}";      EXTRA="" ;;
+  dreamzero)     IMAGE="positro/dreamzero:${IMAGE_TAG}";  EXTRA="" ;;
```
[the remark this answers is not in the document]
vertix: Fixed in 5adce3e: added a `NEBIUS_PRESET` env override (default `1gpu-16vcpu-200gb`), mirroring the `NEBIUS_DISK_SIZE`/`NEBIUS_JOB_TIMEOUT` overrides this same diff introduced. An 8-GPU run now uses `NEBIUS_PRESET=<8-gpu preset>` to match a `*_h100x8` training preset's `torchrun --nproc_per_node=8`. Default stays at one GPU so existing single-GPU jobs are unchanged — this also unblocks gr00t's h100x8 presets, which shared the same hard-coding.

positronic/policy/wrappers.py
vertix: Recommendation: no, keep them in `base.py`. `PolicyWrapper`/`Session`/`DelegatingSession` aren't specific to the serving wrappers — they're the shared contract that `Codec` (`codec.py`) and `_RecordingTap` (`recording.py`) also implement (`Codec(PolicyWrapper)`; both return `Session`/`DelegatingSession`). Moving them into `wrappers.py` would force `codec.py` and `recording.py` to import the concrete-serving module (which pulls in `roboarm`, numpy, deque) just to get the base protocol — inverting the dependency so the foundation lives in a leaf. The current layering is the clean cut: `base.py` = protocol + `|`/`wrap` composition machinery; `wrappers.py`/`codec.py`/`recording.py` = three independent sets of concrete implementers. Happy to revisit if you're picturing a different split (e.g. a thin protocol module separate from both the base helpers and the implementations) — leaving open.

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,173 @@
...
+        while len(self._entries) >= 2 and self._entries[1][0] <= cutoff:
+            self._entries.popleft()
+
+    def sample(self, now: float) -> dict[str, np.ndarray]:
+        times = np.array([t for t, _ in self._entries])
+        picked = [self._entries[self._nearest(times, now + off)][1] for off in self._offsets_sec]
+        return {k: np.stack([frames[k] for frames in picked]) for k in picked[0]}
+
+    @staticmethod
+    def _nearest(times: np.ndarray, target: float) -> int:
```
vertix: It should not be nearest, it should have the same carry over semantic as dataset.Signal have. We should not ever look up in the future, should we?

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,173 @@
...
+
+    Servers whose model conditions on a short video (e.g. DreamZero) need several frames spanning the
+    just-executed chunk at the cadence seen in training, but the harness only forwards an observation
+    to the policy at re-query boundaries. This wrapper sits outside the scheduling wrapper so it sees
+    every control tick: it records each camera frame and substitutes a ``(len(offsets_sec), H, W, 3)``
+    stack sampled at ``offsets_sec`` (ascending negative seconds relative to now).
+    """
+
+    class _Session(DelegatingSession):
+        def __init__(self, inner: Session, image_keys: tuple[str, ...], buffer: _FrameBuffer):
```
vertix: Should not each session own just one buffer? Any pros of sharing buffer between sessions?

positronic/cfg/wrappers.py
```diff
@@ -0,0 +1,18 @@
+import configuronic as cfn
+
+from positronic.policy.wrappers import ChunkedSchedule, ErrorRecovery, TemporalFrameStack
+
+
+@cfn.config()
+def chunked_schedule():
```
vertix: when there's no parameters, it's better to write
error_recovery = cfn.Config(ErrorRecovery)

For ErrorRecovery the same

positronic/vendors/dreamzero/codecs.py
```diff
@@ -236,5 +244,24 @@ def _ik_dreamzero_action(solver: str):
...
+)
+def dreamzero_wrappers(frame_stack, error_recovery, chunked_schedule):
+    """Eval ``wrap`` for DreamZero: AR video context + error recovery + chunked scheduling.
+
+    The frame stack samples the wrist/exterior cameras at ``FRAME_STACK_OFFSETS_SEC`` outside the
+    scheduling wrappers, so the server receives multi-frame context at the trained cadence; pair with
+    the DreamZero codec's full-chunk ``horizon``. Override ``--wrap.chunked_schedule`` to swap the
+    serving schedule (e.g. RTC) while keeping the frame stack.
+    """
+    return frame_stack | error_recovery | chunked_schedule
```
vertix: should not it be error_recovery | frame_stack | chunked_schedule?
Also see no value in frame stack and chunked_schedule arguments right now. frame stack is fine, though

positronic/policy/wrappers.py
vertix: Fixed in a47faf5: replaced the nearest-neighbor pick with `_at_or_before` — `searchsorted(..., side='right') - 1`, so each offset takes the latest frame **at or before** its target time and never a future one (carry-over, same as `dataset.Signal`). Before enough history accumulates it clamps to the oldest frame. Verified: with frames at t=0/2/4 and target 1.4 it now picks t=0, not the nearer t=2.

workflows/nebius/serve.sh
```diff
@@ -64,8 +64,9 @@ case "$VENDOR" in
   # openpi.server imports `openpi_client` at module top → needs --extra openpi
   openpi)        IMAGE="positro/openpi:${IMAGE_TAG}";     EXTRA="--extra openpi " ;;
   gr00t)         IMAGE="positro/gr00t:${IMAGE_TAG}";      EXTRA="" ;;
+  dreamzero)     IMAGE="positro/dreamzero:${IMAGE_TAG}";  EXTRA="" ;;
```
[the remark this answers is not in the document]
vertix: Fixed in a47faf5: added the same `NEBIUS_PRESET` override the train script got (default `1gpu-16vcpu-200gb`). A multi-GPU serve now runs with e.g. `NEBIUS_PRESET=<8-gpu preset>` to match `--num_gpus` / `torchrun --nproc_per_node`. Default is unchanged, so single-GPU endpoints are unaffected.

positronic/policy/wrappers.py
vertix: Agreed — no pros to sharing; a shared buffer would bleed one rollout's frame history into the next. Fixed in a47faf5: `_Session` now constructs its own `_FrameBuffer(offsets_sec)` instead of receiving one, so each session owns exactly one buffer (verified two sessions get independent buffers).

positronic/vendors/dreamzero/codecs.py
vertix: Both done in a47faf5 — flagging my read of the second point so you can confirm:

- **Order**: now `ErrorRecovery() | frame_stack | ChunkedSchedule()`. Matches `default_wrappers` (error recovery outermost) and means a Recover resets the frame buffer through the cancel chain — post-recovery video context restarts clean instead of carrying pre-error frames.
- **Arguments**: dropped `error_recovery` and `chunked_schedule` as `dreamzero_wrappers` args (no real swap point for them today — speculative); kept `frame_stack` as the one arg since it carries the DreamZero offsets. Both stay available standalone as `cfg.wrappers.error_recovery` / `chunked_schedule`. Also removed the now-dead `--wrap.chunked_schedule` note from the README.

Leaving open in case you meant something different by "no value in frame stack and chunked_schedule arguments".

positronic/cfg/wrappers.py
vertix: Done in a47faf5: `chunked_schedule = cfn.Config(ChunkedSchedule)` and `error_recovery = cfn.Config(ErrorRecovery)` (kept `temporal_frame_stack` as `@cfn.config` since it takes params).

Heads-up on a doc tension: `docs/AGENTS.md` currently says "always the `@cfn.config` decorator, never `NAME = cfn.Config(func)`". This case fits as a carve-out — wrapping a zero-arg *class* directly, not a hand-written wrapper function — but the rule reads as absolute. Want me to scope the doc rule to the function case? Leaving open for that call.

vertix: 1. Actually I think we should make temporal_frame_stack just cfn.Config(TemporalFrameStack)
2. Yes, we need to update the rule. Overall the idea is that we should avoid defining a separate function, the only goal of which is to define config, and we should use decorator immediately. When there's a class or function that might be used as is (i.e. without config), then we should use cfn.Config

positronic/vendors/dreamzero/codecs.py
```diff
@@ -7,16 +7,20 @@
...
 from positronic.dataset.transforms.episode import Derive, Get
 from positronic.drivers.roboarm import command
 from positronic.policy.codec import Codec, lerobot_action, lerobot_image, lerobot_state
+from positronic.policy.wrappers import ChunkedSchedule, ErrorRecovery
 
 IMAGE_WIDTH = 320
 IMAGE_HEIGHT = 176
+# Frame offsets (relative seconds) for the AR video context: 4 frames spanning the just-executed
+# 24-step chunk at 15 fps, matching DreamZero's reference eval client schedule [-23, -16, -8, 0].
+FRAME_STACK_OFFSETS_SEC = tuple(o / 15.0 for o in (-23, -16, -8, 0))
```
vertix: 1. Just a simple list is better [-23/15, -16/15, ...]. This is generalisation for the sake of generalisation
2. No need for a constant, let us just put it directly in the _frame_stack arguments as tuple, where it belongs

positronic/vendors/dreamzero/codecs.py
```diff
@@ -214,12 +221,14 @@ def dreamzero_action(tgt_joints_key: str, tgt_grip_key: str, num_joints: int):
     return DreamZeroActionCodec(tgt_joints_key=tgt_joints_key, tgt_grip_key=tgt_grip_key, num_joints=num_joints)
 
 
-# Composed codec: DreamZero observation + action + timing
+# Composed codec: DreamZero observation + action + timing.
+# horizon=None executes the full 24-action chunk DreamZero returns, aligning the re-query with the
+# frame-stack window; a finite horizon (the compose default is 1.0s) would truncate the chunk.
 _action = dreamzero_action.override(tgt_joints_key='robot_command.joints', tgt_grip_key='target_grip')
-joints = codecs.compose.override(obs=dreamzero_obs, action=_action, fps=15.0)
+joints = codecs.compose.override(obs=dreamzero_obs, action=_action, fps=15.0, horizon=None)
```
vertix: Is it really 1 second default? This is weird, I think None should mean "all of it", not 1 sec. Help me to understand

positronic/vendors/dreamzero/codecs.py
vertix: Fixed in 44141d5: dropped the `FRAME_STACK_OFFSETS_SEC` constant and the comprehension — the offsets now sit inline in `_frame_stack` as a tuple `(-23 / 15, -16 / 15, -8 / 15, 0.0)`, with the cadence note as a comment right above it. Also removed the now-dead constant reference from the `dreamzero_wrappers` docstring.

positronic/cfg/wrappers.py
vertix: Both done in 44141d5:
1. `temporal_frame_stack = cfn.Config(TemporalFrameStack)` — the class already coerces `image_keys`/`offsets_sec` to tuples in its constructor, so the wrapper function was pure boilerplate.
2. Updated the rule in `CLAUDE.md`: "never define a function whose only purpose is to build a config — decorate it with `@cfn.config` directly. When wrapping a class or function usable as-is (without configuronic), assign `NAME = cfn.Config(Thing)` rather than writing a wrapper." Left open so you can tweak the wording.

positronic/vendors/dreamzero/codecs.py
vertix: Your model is right: `horizon=None` already means "execute the whole chunk". `compose` only wraps `ActionHorizon` when `horizon is not None` (`cfg/codecs.py:60`), and `ActionHorizon` keeps actions with `timestamp < horizon_sec` — so None = no truncation = all of it. The DreamZero codecs pass `horizon=None`, so they run the full 24-action chunk.

The `1.0` is the *default* on the shared `compose` config (`@cfn.config(fps=15.0, horizon=1.0, ...)`) — "re-query ~1×/sec" for typical chunked policies, not a DreamZero value. I agree a `None`="all" default would read less surprisingly, but flipping `compose`'s default touches every vendor (lerobot/openpi/gr00t) and is out of scope for this PR. Leaving open if you want to track that separately.

vertix: Done in e853c19 (this PR): flipped `compose`'s default to `horizon=None` (= full chunk), and pushed an explicit `horizon=1.0` down to the sites that actually truncate — `act_absolute`, lerobot `ee`/`joints`, lerobot_0_3_3 `ee`/`joints_traj`, and lance's `_compose`/`ee`. The trajectory vendors (gr00t, openpi, dreamzero) dropped their now-redundant `horizon=None` and just use the default. So `None` = "all of it" is the default, and `1.0` is the only explicit horizon left in the tree.

Behavior is unchanged for every existing codec — verified via `action_horizon_sec` meta: ACT codecs still carry 1.0, trajectory codecs carry None. policy/codec tests + golden-pipeline + e2e all green.

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,169 @@
...
+    class _Session(DelegatingSession):
+        def __init__(self, inner: Session, image_keys: tuple[str, ...], offsets_sec: tuple[float, ...]):
+            super().__init__(inner)
+            self._image_keys = image_keys
+            self._buffer = _FrameBuffer(offsets_sec)
+
+        def __call__(self, obs):
+            now = _obs_time(obs)
+            self._buffer.append(now, {k: obs[k] for k in self._image_keys})
+            return self._inner({**obs, **self._buffer.sample(now)})
```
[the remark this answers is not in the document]
vertix: Fixed in c58158a: taught the recorder's `_as_image` to treat a `(T, H, W, 3)` temporal stack as an image and log its most-recent frame (the current observation), instead of letting it fall through to per-pixel numeric series. So a DreamZero server started with `--recording_dir` records one usable RGB frame per camera per inference rather than hundreds of thousands of scalars.

Verified: `(4, H, W, 3)` → most-recent frame logged as an image; `(1, H, W, 3)` and `(H, W, 3)` unchanged; joint/action vectors (`(7,)`, `(24, 8)`) still log as numeric series. `test_recording` green.

positronic/policy/wrappers.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def __init__(self, offsets_sec: tuple[float, ...]):
+        self._offsets_sec = offsets_sec
+        self._entries: deque[tuple[float, dict[str, np.ndarray]]] = deque()
+
+    def reset(self):
+        self._entries.clear()
+
+    def append(self, now: float, frames: dict[str, np.ndarray]):
+        self._entries.append((now, {k: np.array(v) for k, v in frames.items()}))
```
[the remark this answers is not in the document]
vertix: Fixed in c93ba79: `_FrameBuffer.append` now skips storing a frame that's byte-identical to the previous entry. On real DROID (100 Hz harness, hd720@30fps cameras) the buffer was holding ~150 copies/camera over the 1.53 s window, ~2/3 of them re-copies of unchanged frames; it now keeps only the ~46 distinct frames (~3× less). Carry-over sampling is unchanged — `_at_or_before` already reuses the stored entry for the ticks where no new frame arrived, and offset 0 still resolves to the current frame (verified). No `min_dt` reintroduced, so no zero-offset skew.

The buffer still inherently holds ~46 hd720 frames (~250 MiB) to cover the window; cutting that further would mean downsampling before buffering, which couples the wrapper to the model's input resolution — left out of scope.

Issue #433, opened 2026-06-15
Split EvalUI into a reusable review surface and a lifecycle driver


vertix:
## Problem

`EvalUI` (`positronic/gui/eval.py`) currently fuses two concerns that PR #431 made it clear should be separate:

1. **The lifecycle driver** — emits RUN/FINISH/HOME directives, shows the time budget, owns the trial state machine. This is the attended counterpart to the unattended `trials` plan.
2. **The review surface** — navigates recorded episodes, scrubs video, and edits the per-episode static fields (outcome, counts, notes, object, tote/camera placement), persisting each change to the edit log.

The review surface is the more reusable half: it is just a view over a dataset directory plus its edit log, and it would be valuable on its own — ideally the **positronic server** (`positronic/server/`) grows into it, so episodes can be reviewed and corrected from the same place they are inspected, without a live robot attached.

## Why split (a concrete goal)

Today the driver has to know the dataset output directory **only because the review half lives inside it** — the editor reads and writes the edit log. That forces `inference.py` to build drivers as a factory `Callable[[Path | None], Driver]`, deferred until `pos3.sync` resolves the local output root at runtime, instead of constructing them eagerly from config.

Separating the two removes that reason: the **driver** becomes a pure directive source that needs nothing from the run and can be built eagerly again, while the **review surface** is the only thing that takes the resolved output directory. Killing the driver-closure-over-`output_dir` is an explicit goal of this change (it's the subject of the open factory-closure discussion on PR #431).

## Open question

The static fields an operator edits **differ per use case**. The phail eval has `eval.outcome`, `eval.successful_items`, `eval.object`, `eval.tote_placement`, `eval.external_camera`; another task family will want a different set. So the review surface cannot hard-code the phail form — it needs a way to declare "these are the editable fields, with these widgets/validation" per eval (or per `Task`/`Eval`).

A natural anchor is the rig's `Task`/`Eval` (`positronic/eval.py`): the same place that already declares privileged signals and the policy instruction could declare the review schema. But the exact shape (where it lives, how widgets map to field types, how validation like `successful <= total` is expressed) is unresolved.

## Asks

- Decide the boundary between the driver and the review surface, and whether the review surface lives in `positronic/gui/` or moves toward the server.
- Restore eager driver construction once the review surface owns `output_dir` — drop the `Callable[[Path | None], Driver]` factory in `inference.py`.
- Design the per-eval editable-field schema so the review form is data-driven rather than phail-specific.

[...]

PR #435, opened 2026-06-15
Add `real.droid.pick_place` eval family + droid hardware-identity meta

vertix:
## Summary
- New `positronic/cfg/eval/real/droid.py`: the real tote pick-and-place `Eval` family — `pick_place` (unified instruction) plus `pick_place_towels` / `_spoons` / `_scissors` / `_batteries` variants, mirroring `cfg/eval/sim/positronic.py`. The `droid` embodiment is injected as a cfn param; the `Task` ships `privileged={}`, `reset=None` (human reset), no success criterion (outcome is the operator's annotation), and a per-trial `timeout`. Reachable via `positronic eval run --eval=.real.droid.pick_place`.
- New `positronic/cfg/eval/real/tasks.py`: a dependency-free single source of truth for the PhAIL tote pick-place instruction strings, alongside the real eval configs that define them. Previously duplicated across `cfg/ds/internal.py`, `cfg/phail/v1_0.py`, `gui/eval.py`, `cfg/analysis.py`, and `utilities/fake_dataset_generator.py`; all now import from it (repo-wide grep confirms one definition).
- Made `cfg/eval/__init__.py` a light relative-ref anchor: the runtime-dispatching `run` config (and its `inference` import) moved to `cfg/eval/run.py`, so the package init no longer pulls the runtime. This lets the task constants import from `cfg.eval.*` with no `gui.eval` <-> `cfg.eval` import cycle.
- Roadmap step **6b**, part (a). The hardware-identity meta (e) was dropped on review: the robot IP is useless in practice, and per-camera `serial_number` accessors break the uniform camera API — the right home is a `meta` signal on every driver, now captured as roadmap step 14. The attended-console eval-first refit (d) and docs-promotion gate (f) stay deferred with #433.

## Test plan
- `positronic eval run --eval=.real.droid.pick_place --help` resolves the full config tree.
- Single source wired across all six consumers; fresh-interpreter imports of `gui.eval` / `cfg.ds.internal` / `cfg.analysis` / `inference` / `cli` all succeed (no import cycle); ruff clean.
- `uv run --locked pytest --no-cov`: 578 passed, 7 skipped.
- **Manual (hardware-gated, not run here):** a real-droid run on the physical Franka + ZED + Robotiq rig.


positronic/cfg/embodiment.py
```diff
@@ -32,11 +32,16 @@ def droid(robot_arm, gripper, cameras):
         'robot_command': Command(robot_arm.commands, roboarm_command.Reset(), Serializers.robot_command),
         'target_grip': Command(gripper.target_grip, 0.0, None),
     }
+    static_meta = {
+        **ROBOT_STATIC_META,
+        'hardware.robot_ip': robot_arm.ip,
```
vertix: We likely had it in the doc, but in reality it's useless

positronic/cfg/eval/real/droid.py
```diff
@@ -0,0 +1,27 @@
+import configuronic as cfn
+
+from positronic.cfg.embodiment import droid
+from positronic.eval import Eval, Task
+
+UNIFIED_TASK = 'Pick all the items one by one from transparent tote and place them into the large grey tote.'
+TOWELS_TASK = 'Pick all the towels one by one from transparent tote and place them into the large grey tote.'
+SPOONS_TASK = 'Pick all the wooden spoons one by one from transparent tote and place them into the large grey tote.'
+SCISSORS_TASK = 'Pick all the scissors one by one from transparent tote and place them into the large grey tote.'
```
vertix: We also had batteries

positronic/drivers/camera/zed.py
```diff
@@ -64,6 +64,10 @@ def __init__(
         self.depth_mask: pimm.SignalEmitter = pimm.ControlSystemEmitter(self)
         self._depth_mask_adapter = None  # Lazy init
 
+    @property
+    def serial_number(self) -> int | None:
```
vertix: This is a bad idea. The reason is that all cameras must have the same API, so we can switch easy from one to another.
The real solution is smth like robot meta that we have for robots. Ready to discuss more offline

positronic/cfg/embodiment.py
vertix: Fixed in 42846f5: dropped `hardware.robot_ip` — and reverted the whole hardware-identity meta block with it, so `droid` is back to `dict(ROBOT_STATIC_META)`. The arm's measured `robot_meta` already rides into static via `meta_source`, so robot identity isn't lost.

positronic/cfg/eval/real/droid.py
vertix: Fixed in 42846f5: added `BATTERIES_TASK` + `pick_place_batteries` (mirrors `cfg/ds/internal.py`'s `ALL_TASKS`).

positronic/drivers/camera/zed.py
vertix: Agreed. Reverted the `serial_number` accessor in 42846f5 (the camera-serial stamping in `droid` went with it), so the uniform camera API is intact. A camera-meta emitter analogous to `robot_meta` is the right home for serials — leaving this open for the offline discussion on that design.

positronic/cfg/eval/real/droid.py
```diff
@@ -0,0 +1,29 @@
+import configuronic as cfn
+
+from positronic.cfg.embodiment import droid
+from positronic.eval import Eval, Task
+
+UNIFIED_TASK = 'Pick all the items one by one from transparent tote and place them into the large grey tote.'
```
vertix: I think we have these string constants already in this repo, only one should stay

vertix: Fixed in e9a0727: hoisted all the PhAIL tote pick-place instructions into a new dependency-free `positronic/cfg/tasks.py` as the single source, and pointed every copy at it — `cfg/eval/real/droid.py`, `cfg/ds/internal.py`, `cfg/phail/v1_0.py`, `gui/eval.py`, `cfg/analysis.py` (its scoring `match`), and `utilities/fake_dataset_generator.py`. A repo-wide grep confirms the strings are now defined exactly once. Kept the module dep-free so the eval CLI doesn't drag in `positronic_server`, which all the prior homes do.

vertix: Follow-up (409e8a1): relocated the single source from `cfg/tasks.py` to `cfg/eval/real/tasks.py` per offline feedback — these are the real-eval task definitions, so they belong with the eval configs. That required lazy-importing `inference` in `cfg.eval.run` to break the `gui.eval` <-> `cfg.eval` import cycle the move introduced (a config package importing the runtime at load time).

vertix:
@codex review

positronic/cli/eval/run.py
```diff
@@ -0,0 +1,146 @@
...
+            'eval.trial_index': i,
+            'eval.trial_count': trial_count,
+        }
+        for i in range(trial_count)
+    ]
+    main(
+        embodiment=eval.embodiment,
+        task=eval.task,
+        policy=policy,
+        trials=trials,
```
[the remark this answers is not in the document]
vertix: Correct observation, and it's the known, documented gap rather than a regression. This PR intentionally ships only part (a) of roadmap step 6b — the real `Eval` family as the **unattended** substrate (`positronic eval run --eval=.real.droid.pick_place` → self-driving `trials`, timeout-terminated, machine metadata only).

Wiring the operator console (EvalUI: `eval.outcome`, item counts, object/tote/camera, drop/undrop) to *consume* these eval configs is part **(d)** of 6b — the attended-console eval-first refit — which is deferred together with #433 (the EvalUI review-surface/driver split). It did not land here on purpose; the PR body and the design doc both record (d)/(f) as deferred.

Until (d) lands, real evals are still run **attended** via the unchanged `positronic-inference phail` flow (same `droid` embodiment + `eval_ui` driver), so operator annotations are collected there. So nothing regresses; `--eval=.real.droid.pick_place` just isn't the operator-facing entry yet.

Leaving this open as a tracked follow-up (6b-d / #433), not resolving.

PR #437, opened 2026-06-16
Document the DreamZero wan2.2 convert→train→serve pipeline

vertix:
## Summary
- Rewrite `positronic/vendors/dreamzero/README.md` to document the full **convert → train → serve → infer → view** pipeline, so the wan2.2 fine-tune flow is reproducible end-to-end.
- Pivot the doc to **wan2.2 (5B) primary**, keeping wan2.1 (14B) as a labeled variant; make `--backbone` (and the requirement that it match between train and serve) explicit.
- Add a **Training** section (presets table, fresh + warm-start commands, the `h100x1` `save_only_model`/non-resumable caveat) and a **serve** section covering both the Nebius serverless and docker-compose paths with `--model_path` + `--backbone`.
- Complete the **codecs** table to all four (`joints`, `joints_traj`, `joints_ik`, `joints_ik_sim`) and note inference decode is identical across them.
- Fix stale wan2.1-only facts: training hardware (single H100 works, not "4–8×"), image dims (320×160 for wan2.2), VRAM, and DiT-cache being wan2.1-only.

The convert step is anchored on the dataset's `run_metadata` (codec `joints_ik_sim`, output `s3://interim/sim_stack/dreamzero/joints_ik/`), but uses the **current** config path `@positronic.cfg.ds.sim.sim_stack_cubes` — the metadata's `cfg.ds.phail` path has since moved, so the doc is runnable rather than literal.

## Test plan
Docs-only change (no code). The documented commands were exercised in practice this session: convert command verified against the dataset `run_metadata`; train/serve/infer/view commands match runs executed against `checkpoint-15000`; the dataset config path was confirmed to resolve (`@positronic.cfg.ds.sim.sim_stack_cubes`).

### DIFF
--- a/positronic/vendors/dreamzero/README.md
+++ b/positronic/vendors/dreamzero/README.md
@@ -23,9 +23,9 @@ Pick the backbone with `--backbone` at both train and serve time; **it must matc
 | **Inference** | 1× H100 (80GB) | wan2.2 (5B) ≈ half the VRAM of wan2.1 (14B, ~52GB bf16) |
 | **Training** | 1× or 8× H100 | Full fine-tune or LoRA, DeepSpeed ZeRO-2. Single-H100 full fine-tune works (see presets) |
 
-Train and serve run on an H100 box you reach through a Docker context (see
-[`docker/CONTEXTS.md`](../../../docker/CONTEXTS.md)). The examples below call it `--context <h100>` —
-substitute your own context name. (The Nebius serverless alternative is in the [Appendix](#appendix-nebius-serverless).)
+Train and serve run on an H100 box you reach through a Docker context. The examples below call it
+`--context <h100>` — substitute your own context name. (The Nebius serverless alternative is in the
+[Appendix](#appendix-nebius-serverless).)
 
 ## Zero-shot inference (wan2.1 DROID checkpoint)
 
@@ -58,14 +58,13 @@ the Docker Compose services in [`docker/docker-compose.yml`](../../../docker/doc
 H100 box. Run the `docker compose` commands from the `docker/` directory.
 
 **Prerequisites:**
-- An H100 box reachable via a Docker context ([`docker/CONTEXTS.md`](../../../docker/CONTEXTS.md)); set
-  `CACHE_ROOT` to that box's home so the `~/.cache` and `~/.aws` mounts resolve (S3 credentials come from
-  the mounted `~/.aws`).
-- **Images** are pulled pre-built and do **not** mount local source, so a code change needs a rebuild +
-  push first: convert runs on `positro/positronic` (rebuild with `make push-training`); train and serve run
-  on `positro/dreamzero` (rebuild with `make push-dreamzero-base` then `make push-dreamzero`). A change to
-  `codecs.py` touches both. CI pushes `:latest` and the `main` commit SHA; pin a SHA via `IMAGE_TAG`
-  (e.g. `IMAGE_TAG=dc5e837`) for a reproducible run. See [`docker/Makefile`](../../../docker/Makefile).
+- An H100 box reachable via a Docker context; set `CACHE_ROOT` to that box's home so the `~/.cache` and
+  `~/.aws` mounts resolve (S3 credentials come from the mounted `~/.aws`).
+- **Images**: the commands pull the public prebuilt images by default — `positro/positronic` (convert) and
+  `positro/dreamzero` (train/serve) at the production `:latest` tag — so there is nothing to build. Only if
+  you have **local repo changes** to test do you build and push your own image and pass `IMAGE_TAG=<your-tag>`;
+  otherwise omit it. A `codecs.py` change affects both images. (Maintainer build targets:
+  [`docker/Makefile`](../../../docker/Makefile).)
 - The `s3://interim/…` / `s3://checkpoints/…` paths below are Positronic-internal buckets — substitute your own.
 
 ### 1. Convert the dataset
@@ -87,6 +86,10 @@ CACHE_ROOT=/home/vertix docker --context <h100> compose run --rm lerobot-0_3_3-c
 `dreamzero-train` runs `python -m positronic.vendors.dreamzero.train`, streaming the dataset from
 `--input_path`. Presets (defined in [`train.py`](./train.py)) fix backbone + architecture + GPU count:
 
+> **Only `wan22_full` (wan2.2 full fine-tune) is validated** — that is the configuration these docs were
+> built from. We did **not** get LoRA working, and the `wan2.1` presets are untested here. The other presets
+> exist in code but should be treated as unverified.
+
 | Preset | Backbone | Arch | GPUs | Resumable |
 |--------|----------|------|------|-----------|
 | `wan22_full_h100x1` | wan2.2 (5B) | full | 1 | No — warm-start to continue |
@@ -99,7 +102,7 @@ CACHE_ROOT=/home/vertix docker --context <h100> compose run --rm lerobot-0_3_3-c
 
 ```bash
 cd docker
-CACHE_ROOT=/home/vertix IMAGE_TAG=dc5e837 docker --context <h100> compose run --rm dreamzero-train \
+CACHE_ROOT=/home/vertix docker --context <h100> compose run --rm dreamzero-train \
   wan22_full_h100x1 \
   --input_path=s3://interim/sim_stack/dreamzero/joints_ik \
   --output_path=s3://checkpoints/sim_stack/dreamzero/ \
@@ -128,7 +131,7 @@ API on `8000`:
 
 ```bash
 cd docker
-CACHE_ROOT=/home/vertix IMAGE_TAG=dc5e837 docker --context <h100> compose run --rm --service-ports dreamzero-server \
+CACHE_ROOT=/home/vertix docker --context <h100> compose run --rm --service-ports dreamzero-server \
   --model_path=s3://checkpoints/sim_stack/dreamzero/<exp_name>/checkpoint-<step> \
   --backbone=wan2.2
 ```
@@ -190,8 +193,8 @@ that decodes to a `JointPosition` command. They differ only in how **training la
 If you run on Nebius serverless instead of your own H100 box, the **same recipe** (same presets, codec, and
 args) goes through the `workflows/nebius/*.sh` wrappers — Jobs for convert/train, an Endpoint for serving.
 See [`workflows/nebius/README.md`](../../../workflows/nebius/README.md) for the one-time setup (Nebius CLI
-auth, S3 / MysteryBox credentials, the shared cache filesystem) and `NEBIUS_*` overrides; pin the image with
-`NEBIUS_IMAGE_TAG=<sha>`.
+auth, S3 / MysteryBox credentials, the shared cache filesystem) and `NEBIUS_*` overrides. It uses the
+production image by default; set `NEBIUS_IMAGE_TAG=<tag>` only to test a specific build.
 
 ```bash
 # Convert (CPU job)
@@ -201,7 +204,7 @@ bash workflows/nebius/convert.sh lerobot_0_3_3 \
   --output_dir=s3://interim/sim_stack/dreamzero/joints_ik/
 
 # Train (H100 job; ~3 days for 30k steps on one H100 — size the timeout accordingly)
-NEBIUS_IMAGE_TAG=dc5e837 NEBIUS_JOB_TIMEOUT=96h \
+NEBIUS_JOB_TIMEOUT=96h \
 bash workflows/nebius/train.sh dreamzero wan22_full_h100x1 \
   --input_path=s3://interim/sim_stack/dreamzero/joints_ik \
   --output_path=s3://checkpoints/sim_stack/dreamzero/ \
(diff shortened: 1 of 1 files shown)


positronic/vendors/dreamzero/README.md
```diff
@@ -1,58 +1,197 @@
...
+This is Positronic's standard [convert → train → serve → infer](../../../docs/training-workflow.md) loop.
+Convert runs on CPU; train and serve run on H100. The cloud wrappers live in `workflows/nebius/` — see
+[`workflows/nebius/README.md`](../../../workflows/nebius/README.md) for the one-time setup it assumes
+(Nebius CLI auth, S3 / MysteryBox credentials, the shared cache filesystem) and the `NEBIUS_*` overrides.
+The `s3://interim/…` and `s3://checkpoints/…` paths below are Positronic-internal buckets — substitute
+your own.
+
+**Image:** the wrappers pull `positro/dreamzero:${NEBIUS_IMAGE_TAG:-latest}`, which does not mount local
+source, so a code change needs a rebuild + push first — `make push-dreamzero-base` then
+`make push-dreamzero IMAGE_TAG=<tag>` (see [`docker/README.md`](../../../docker/README.md) /
```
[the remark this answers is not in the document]
vertix: Fixed in a018740. The README was restructured so the dockerized pipeline is primary, and the image guidance now distinguishes the two images per step: **convert** runs on `positro/positronic` (the `lerobot-0_3_3-convert` compose service / `make push-training`), while **train and serve** run on `positro/dreamzero`. Noted that a `codecs.py` change touches both. Good catch — thanks.

positronic/vendors/dreamzero/README.md
```diff
@@ -1,58 +1,178 @@
...
+### 1. Convert the dataset
+
+The DreamZero codecs reuse the shared `lerobot_0_3_3` converter. The sim stack-cubes dataset is
+converted with the `joints_ik_sim` codec (joint targets reconstructed from recorded EE-pose targets
+via the `dm_control` IK solver):
+
+```bash
+bash workflows/nebius/convert.sh lerobot_0_3_3 \
+  --dataset.dataset=@positronic.cfg.ds.sim.sim_stack_cubes \
+  --dataset.codec=@positronic.vendors.dreamzero.codecs.joints_ik_sim \
```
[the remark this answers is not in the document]
vertix: Declining — this is the recipe we ran end-to-end. The dataset was converted with `joints_ik_sim` (no `image_size` override), then trained with `wan22_full_h100x1` and served + sim-evaluated against `checkpoint-15000`, all fine.

The wan2.2 input resolution is set by the trainer, not the codec: `train.py` passes `image_resolution_width=320` / `image_resolution_height=160` to the Hydra experiment, so frames are resized to 160 at load regardless of the codec's intermediate `(320, 176)`. Adding a per-backbone codec `image_size` override would document an untested variant. I added a note to Technical Details (a018740) spelling this out. Exposing a wan2.2-native size on the codec would be a separate code change; leaving this open.

positronic/vendors/dreamzero/README.md
```diff
@@ -1,58 +1,196 @@
...
+  --backbone=wan2.2
+```
+
+Sanity-check once warm: `curl http://<host>:8000/api/v1/models` → `{"models": ["<model_path>"]}`.
+
+### 4. Run sim inference
+
+```bash
+uv run --locked positronic-inference sim \
+  --policy=.remote --policy.host=<host> --policy.port=8000 \
```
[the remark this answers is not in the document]
vertix: Declining — the Nebius serverless endpoint is exposed on `:8000`, not 80. `serve.sh` creates it with `--container-port 8000 --public`, and `public_endpoints[0]` comes back as `<IP>:8000`, so the banner's `Endpoint URL` / `curl http://$IP/...` already carry `:8000`. Verified in this PR's runs: I served `checkpoint-15000` this way and ran sim inference against it with `--policy.host=<IP> --policy.port=8000` (7 episodes) with no connectivity issue — so `8000` is correct for both the docker-compose and Nebius paths. (The Nebius commands now live in the Appendix after the a018740 restructure.) Leaving this open.

positronic/vendors/dreamzero/README.md
```diff
@@ -1,58 +1,217 @@
...
+
+## Full pipeline (fine-tune your own checkpoint)
 
-## Quick Start
+The end-to-end loop — [convert → train → serve → infer](../../../docs/training-workflow.md) — runs through
+the Docker Compose services in [`docker/docker-compose.yml`](../../../docker/docker-compose.yml) on your
+H100 box. Run the `docker compose` commands from the `docker/` directory.
+
+**Prerequisites:**
+- An H100 box reachable via a Docker context ([`docker/CONTEXTS.md`](../../../docker/CONTEXTS.md)); set
```
vertix: Don't link this file

positronic/vendors/dreamzero/README.md
```diff
@@ -1,58 +1,217 @@
...
+  `CACHE_ROOT` to that box's home so the `~/.cache` and `~/.aws` mounts resolve (S3 credentials come from
+  the mounted `~/.aws`).
+- **Images** are pulled pre-built and do **not** mount local source, so a code change needs a rebuild +
```
vertix: External people don't have access to pushing images. Just say that if you have local changes in the repo, you need to use your own image, otherwise you can use public positornic image

positronic/vendors/dreamzero/README.md
```diff
@@ -1,58 +1,217 @@
...
+cd docker
+CACHE_ROOT=/home/vertix docker --context <h100> compose run --rm lerobot-0_3_3-convert convert \
+  --dataset.dataset=@positronic.cfg.ds.sim.sim_stack_cubes \
+  --dataset.codec=@positronic.vendors.dreamzero.codecs.joints_ik_sim \
+  --output_dir=s3://interim/sim_stack/dreamzero/joints_ik/
+```
+
+### 2. Train
+
+`dreamzero-train` runs `python -m positronic.vendors.dreamzero.train`, streaming the dataset from
```
vertix, line 90: We should state clearly in this section, that finetuning only worked for 2.2 full-finetune, that we did not manage to get LoRA working.

positronic/vendors/dreamzero/README.md
```diff
@@ -1,58 +1,217 @@
...
+| `wan22_lora_h100x1` | wan2.2 (5B) | LoRA | 1 | LoRA trains from zero |
+| `wan22_lora_h100x8` | wan2.2 (5B) | LoRA | 8 | Yes (optimizer restored) |
+| `wan21_*` | wan2.1 (14B) | full / LoRA | 1 / 8 | As above |
+
+**Fresh run** (from the base backbone):
+
+```bash
 cd docker
-CACHE_ROOT=/home/vertix docker --context vm-dreamzero compose run --rm --service-ports dreamzero-server
+CACHE_ROOT=/home/vertix IMAGE_TAG=dc5e837 docker --context <h100> compose run --rm dreamzero-train \
```
vertix: Why this IMAGE_TAG? I know it is tested, but this will become stale very soon. One should use the production tag, that, I suppose, is just ommitting this env var

Here and everywhere

positronic/vendors/dreamzero/README.md
```diff
@@ -1,58 +1,217 @@
...
+  wan22_full_h100x1 \
+  --input_path=s3://interim/sim_stack/dreamzero/joints_ik \
+  --output_path=s3://checkpoints/sim_stack/dreamzero/ \
+  --exp_name=<exp_name> \
+  --max_steps=30000 --save_steps=2500 --gradient_accumulation_steps=4 --save_total_limit=9999
+```
+
+Checkpoints land at `s3://checkpoints/sim_stack/dreamzero/<exp_name>/checkpoint-<step>`.
+
+**Warm-start** from a prior run's weights (fresh optimizer + schedule) — the only way to extend an
```
vertix: This is impossible to understand. Just say that because of the internal details, the DreamZero training does not save optimisation state, so when we resume, Adam starts fresh an we lose the state

positronic/vendors/dreamzero/README.md
```diff
@@ -1,58 +1,217 @@
...
+| `joints_traj` | Recorded state (`robot_state.q`) + grip | Trajectory / executed-state targets |
+| `joints_ik` | Joints solved from recorded EE-pose targets via IK (`dls_limits` solver) | EE-driven datasets |
+| `joints_ik_sim` | `joints_ik` with the `dm_control` IK solver | Sim datasets (used for `sim_stack_cubes`) |
 
 ## Technical Details
 
-- **Action space**: Absolute joint positions (7 DoF) + gripper (1)
-- **Observation**: 3 cameras (2 exterior + 1 wrist) at 320×176 + joint state + language prompt
-- **Action horizon**: 24 timesteps per inference
+- **Action space**: absolute joint positions (7 DoF) + gripper (1)
```
vertix: Is it really absolute? I was thinking it is relative, but pls check

positronic/vendors/dreamzero/README.md
```diff
@@ -1,58 +1,217 @@
...
+- **Observation**: 3 cameras (2 exterior + 1 wrist) + joint state + language prompt; 320×160 (wan2.2),
+  320×176 (wan2.1). The wan2.2 trainer sets `image_resolution_height=160` and resizes at load, so the
+  codec's intermediate `(320, 176)` does not need a per-backbone override.
+- **Action horizon**: 24 timesteps per inference; the `dreamzero_wrappers` re-query aligns the
+  chunk schedule with the AR frame-stack window
 - **Wire protocol**: roboarena WebSocket + msgpack-numpy
```
vertix: Wire protocol is our standard Positronic wire protocol, you can link it

positronic/vendors/dreamzero/README.md
```diff
@@ -1,58 +1,217 @@
...
-- **Inference latency**: ~5s on 1× H100 (with DiT caching)
-- **No fork needed**: Uses Hydra YAML configs, injectable from outside
+- **DiT caching**: wan2.1 only (skipped for wan2.2)
+- **No Positronic fork**: upstream DreamZero is used unmodified (pinned SHA in [`Dockerfile`](./Dockerfile));
```
vertix, line 193: This is too big detail, omit

positronic/vendors/dreamzero/README.md
vertix: Done in 2257bdc — dropped both links to `docker/CONTEXTS.md`; the doc now just says "an H100 box reachable via a Docker context."

positronic/vendors/dreamzero/README.md
vertix: Done in 2257bdc — the Images note now says the commands use the public `positro/*` images by default (nothing to build). You only build/push your own image and set `IMAGE_TAG` if you have local repo changes to test; otherwise omit it.

positronic/vendors/dreamzero/README.md
vertix: Done in 2257bdc — added a callout at the top of the Train section: only `wan22_full` (wan2.2 full fine-tune) is validated; we did not get LoRA working, and the wan2.1 presets are untested.

positronic/vendors/dreamzero/README.md
vertix: Done in 2257bdc — removed the pinned `IMAGE_TAG=dc5e837` from every command (train, serve, and the Nebius appendix's `NEBIUS_IMAGE_TAG`). They now use the production image by default (omit the env var); the var is documented only for testing a specific build. Applied everywhere, as you noted.

positronic/vendors/dreamzero/README.md
```diff
@@ -1,58 +1,217 @@
...
+
+The DreamZero codecs reuse the shared `lerobot_0_3_3` converter (a CPU step — run it on any box, or locally
+with `uv run`). The sim stack-cubes dataset is converted with the `joints_ik_sim` codec (joint targets
+reconstructed from recorded EE-pose targets via the `dm_control` IK solver):
 
 ```bash
-# 1. Start an H100 VM
-../internal/scripts/start.sh dreamzero
+cd docker
+CACHE_ROOT=/home/vertix docker --context <h100> compose run --rm lerobot-0_3_3-convert convert \
```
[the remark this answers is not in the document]
vertix: Addressed in 2257bdc, the other way around per maintainer preference: rather than pin the convert tag, the pins were removed from all stages, so convert/train/serve consistently use the production `:latest` image — one revision across the runbook, without a tag that goes stale. The `IMAGE_TAG` / `NEBIUS_IMAGE_TAG` override stays documented for testing a specific build.

positronic/vendors/dreamzero/README.md
vertix: Done in 5ac3580 — simplified to: the DreamZero trainer doesn't persist optimizer state, so resuming restarts Adam from scratch; warm-start loads the weights and accepts a fresh optimizer + schedule. Dropped the `save_only_model` / DeepSpeed jargon.

positronic/vendors/dreamzero/README.md
vertix: You're right — I was describing only the decoded boundary. Checked upstream: `groot/vla/configs/data/dreamzero/droid_relative_wan22.yaml` sets `relative_action: true` with `relative_action_keys: [joint_position]`, so DreamZero predicts the joints **relative to the current state** internally; the absolute `JointPosition` is reconstructed at serve time. Corrected in d55b3bb.

positronic/vendors/dreamzero/README.md
vertix: Done in 5ac3580 — now reads "Positronic's standard WebSocket protocol" linking `docs/connect-your-model.md`, and dropped the roboarena/msgpack-numpy low-level detail.

positronic/vendors/dreamzero/README.md
vertix: Done in 5ac3580 — removed the DiT-caching bullet.

PR #438, opened 2026-06-17
Remote env-server boundary (`RemoteEnvControlSystem` + `EnvAdapter`)

vertix:
## Summary

Second of two stacked PRs (stacks on #448 — the harness / control-loop refactor). Adds the remote env-server boundary.

- New `positronic/simulator/env_server/` module: a socket `proxy`/`client`/`protocol`, a `server`, and an `EnvAdapter` translating env steps to and from the wire. `RemoteEnvControlSystem` stays a dumb translator; the adapter owns trajectory playing and the canonical↔raw mappings.
- Reuses the harness's frame-0 / reset-as-flag / record-ordering contract across the socket boundary — the shared harness+recorder invariants are proven locally in #448.
- Tests cover the boundary itself: wire round-trip parity, the proxy's run loop (frame-0 + free-run), the real `stack_cubes` wrapper end-to-end over the socket, and error-frame propagation.

## Test plan

`uv run --locked pytest --no-cov` → 593 passed, 7 skipped (full suite on the stacked branch, including the 4 env-server boundary tests).

### DIFF
--- a/positronic/simulator/env_server/client.py
+++ b/positronic/simulator/env_server/client.py
@@ -1,0 +1,55 @@
+"""Synchronous client for the env server: the lockstep ``reset``/``step``/``close`` round-trips.
+
+Positronic-free (``websockets`` + the wire codec). ``RemoteEnvControlSystem`` wraps this as a pimm
+control system; tests use it directly to compare a socket rollout against an in-process one.
+"""
+
+import time
+from typing import Any
+
+from websockets.sync.client import connect
+
+from .protocol import decode, encode
+
+
+class EnvConnection:
+    """One websocket to an ``EnvServer``, opened with retry. ``reset``/``step`` block on the round-trip.
+
+    There is no handshake: the first ``reset`` constructs the env server-side and returns
+    ``{'obs', 'meta', 'control_dt'}``.
+    """
+
+    def __init__(self, host: str, port: int, *, open_timeout: float = 10.0, connect_deadline: float = 60.0):
+        uri = f'ws://{host}:{port}/'
+        deadline = time.monotonic() + connect_deadline
+        backoff = 0.5
+        while True:
+            try:
+                # Camera + full-state observations routinely exceed websockets' 1 MiB default frame size.
+                self._ws = connect(uri, open_timeout=open_timeout, max_size=None)
+                break
+            except (TimeoutError, OSError) as e:
+                if time.monotonic() >= deadline:
+                    raise type(e)(f'{e} (connecting to {host}:{port})') from e
+                time.sleep(backoff)
+                backoff = min(backoff * 2, 5.0)
+
+    def reset(self, token: Any) -> dict[str, Any]:
+        return self._request({'cmd': 'reset', 'token': token})
+
+    def step(self, action: dict[str, Any]) -> dict[str, Any]:
+        return self._request({'cmd': 'step', 'action': action})
+
+    def _request(self, msg: dict[str, Any]) -> dict[str, Any]:
+        self._ws.send(encode(msg))
+        result = decode(self._ws.recv())
+        if 'error' in result:
+            raise RuntimeError(f'env server: {result["error"]}')
+        return result
+
+    def close(self) -> None:
+        try:
+            self._ws.send(encode({'cmd': 'close'}))
+            self._ws.recv()
+        finally:
+            self._ws.close()
--- a/positronic/simulator/env_server/protocol.py
+++ b/positronic/simulator/env_server/protocol.py
@@ -1,0 +1,39 @@
+"""Wire codec for the remote env-server boundary: msgpack with a numpy envelope.
+
+This module is **positronic-free** — it imports only ``msgpack`` and ``numpy`` — so it can be
+imported (or copied) into a benchmark's isolated interpreter alongside the dumb server without
+dragging in pimm or the rest of positronic. Only raw numpy arrays and plain-data dicts cross the
+wire; every canonical<->raw mapping lives client-side in the ``EnvAdapter``.
+
+Arrays travel as raw bytes plus their ``dtype.str`` and shape, so a numpy-2 server round-trips a
+numpy-1 client unchanged.
+"""
+
+import functools
+
+import msgpack
+import numpy as np
+
+
+def _pack(obj):
+    if isinstance(obj, np.ndarray):
+        if obj.dtype.kind in ('V', 'O', 'c'):
+            raise ValueError(f'Unsupported dtype: {obj.dtype}')
+        return {b'__ndarray__': True, b'data': obj.tobytes(), b'dtype': obj.dtype.str, b'shape': obj.shape}
+    if isinstance(obj, np.generic):
+        return {b'__npgeneric__': True, b'data': obj.item(), b'dtype': obj.dtype.str}
+    return obj
+
+
+def _unpack(obj):
+    if b'__ndarray__' in obj:
+        # ``bytearray`` (not the raw msgpack ``bytes``) backs a writable array, so the socket path
+        # matches the in-process path for envs/adapters that mutate a decoded buffer in place.
+        return np.ndarray(buffer=bytearray(obj[b'data']), dtype=np.dtype(obj[b'dtype']), shape=obj[b'shape'])
+    if b'__npgeneric__' in obj:
+        return np.dtype(obj[b'dtype']).type(obj[b'data'])
+    return obj
+
+
+encode = functools.partial(msgpack.packb, default=_pack)
+decode = functools.partial(msgpack.unpackb, object_hook=_unpack)
--- a/positronic/simulator/env_server/proxy.py
+++ b/positronic/simulator/env_server/proxy.py
@@ -1,0 +1,90 @@
+"""``RemoteEnvControlSystem``: a remote env server driven as one pimm control system.
+
+A dumb translator: it owns the pimm ports (command receivers, observation + privileged emitters,
+``robot_meta``, the Task's ``done``) and the trial lifecycle, but no command logic. Each control period
+it hands the latest command messages to the ``EnvAdapter``, round-trips the raw action it returns over
+the wire, and re-emits the canonical signals the adapter maps back — so only raw arrays cross the
+boundary and the World's virtual clock advances by the env's ``control_dt`` per step. The adapter owns
+trajectory playing, holding, and the canonical<->raw mappings; ``control_dt`` is whatever the latest
+observation reports (``reset`` and every ``step``).
+"""
+
+from collections.abc import Iterator
+from typing import Any
+
+import pimm
+from positronic.simulator.env_server.adapter import EnvAdapter
+from positronic.simulator.env_server.client import EnvConnection
+
+# Pacing before the first reset, when the env's ``control_dt`` is still unknown. Only sets the instant
+# frame-0 lands at, then the env's reported ``control_dt`` takes over.
+_IDLE_DT = 0.1
+
+
+class RemoteEnvControlSystem(pimm.ControlSystem):
+    def __init__(self, adapter: EnvAdapter, host: str, port: int):
+        self._adapter = adapter
+        self._conn = EnvConnection(host, port)
+
+        self.commands: pimm.ReceiverDict = pimm.ReceiverDict(self, default=None)
+        self.observations: pimm.EmitterDict = pimm.EmitterDict(self)
+        self.privileged: pimm.EmitterDict = pimm.EmitterDict(self)
+        self.robot_meta: pimm.SignalEmitter = pimm.ControlSystemEmitter(self)
+        self.done: pimm.SignalEmitter = pimm.ControlSystemEmitter(self)
+
+        # A trial is live between reset and the env's done. The proxy steps only then — not before the
+        # first reset (Gym envs reject step-before-reset), not after done. It sleeps every turn regardless.
+        self._active = False
+        # The latest env frame (``obs`` + ``control_dt``), refreshed by ``reset`` and each ``step``.
+        self._frame: dict[str, Any] | None = None
+        # Set by ``reset``; the run loop publishes frame-0 (instead of stepping) on its next turn and clears it.
+        self._reset_pending = False
+
+    def reset(self, seed: int | None = None) -> None:
+        """Re-randomize the env from ``seed`` and arm frame-0 publication for the next turn (the ``RUN`` hook).
+
+        Resets the remote env (acquiring the fresh frame and its ``control_dt``), then flags the run loop
+        to publish the scene meta, a full observation payload (frame-0) and a non-terminal ``done`` on its
+        next turn — in sequence, so the recorder samples frame-0 before any step. Stale commands queued
+        while inactive (e.g. the inter-episode home) are dropped so the first step doesn't apply them.
+        """
+        for _, receiver in self.commands.items():
+            receiver.read()
+        self._frame = self._conn.reset(self._adapter.reset_token(seed))
+        self._reset_pending = True
+        self._active = True
+
+    def _emit_payload(self, raw_obs: dict[str, Any]) -> None:
+        for name, value in self._adapter.observations(raw_obs).items():
+            self.observations[name].emit(value)
+        for name, value in self._adapter.privileged(raw_obs).items():
+            self.privileged[name].emit(value)
+
+    def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[pimm.Command]:
+        try:
+            while not should_stop.value:
+                # The proxy is the eval's sole time-master: it sleeps one control period every turn —
+                # stepping, publishing frame-0, or idle between trials alike. Before the first reset the
+                # env's ``control_dt`` is unknown, so it paces at ``_IDLE_DT`` until reset reports the real one.
+                yield pimm.Sleep(self._frame['control_dt'] if self._frame is not None else _IDLE_DT)
+                if self._reset_pending:
+                    # The reset is this turn's step: publish the env's frame-0 (no step) and clear the prior
+                    # terminal, so the recorder samples it before any step advances the env.
+                    self._reset_pending = False
+                    self.robot_meta.emit(self._frame['meta'])
+                    self._emit_payload(self._frame['obs'])
+                    self.done.emit({})
+                elif self._active:
+                    self._frame = self._step_env(clock)
+                    self._emit_payload(self._frame['obs'])
+        finally:
+            self._conn.close()
+
+    def _step_env(self, clock: pimm.Clock) -> dict[str, Any]:
+        commands = {name: receiver.read() for name, receiver in self.commands.items()}
+        result = self._conn.step(self._adapter.action(commands, clock.now_ns()))
+        payload = self._adapter.terminal(result)
+        if payload:  # truthy-valued done: a non-empty payload ends the trial, an empty/``None`` one continues
+            self.done.emit(payload)
+            self._active = False
+        return result
--- a/positronic/simulator/env_server/server.py
+++ b/positronic/simulator/env_server/server.py
@@ -1,0 +1,109 @@
+"""The dumb remote env-server: a benchmark env behind ``reset``/``step``/``close`` over websockets.
+
+Positronic-free by contract — it depends only on ``websockets``, plus the wire codec and the
+``EnvProtocol`` an env implements. A benchmark runs this in its own isolated interpreter (where
+positronic can't be installed); the native fixture runs it in-process against ``MujocoSim``. The
+server is lockstep request-response — one client, one outstanding request — so the World's virtual
+clock advances unchanged while a step round-trips.
+
+There is no build phase: ``reset`` constructs (or reuses a cached) env from its opaque token and
+re-randomizes it; ``control_dt`` rides every observation (``reset`` and each ``step``), so a benchmark
+may even vary its control period per step. Heavy construction is the env's own concern (cache it).
+
+Protocol (msgpack frames, see ``protocol``):
+  client -> ``{'cmd': 'reset', 'token': ...}``    server -> ``{'obs': {...}, 'meta': {...}, 'control_dt': float}``
+  client -> ``{'cmd': 'step', 'action': {...}}``   server -> ``{'obs': {...}, 'done': bool, 'control_dt': float}``
+  client -> ``{'cmd': 'close'}``                    server -> ``{'ok': True}``
+Any command whose handling raises returns ``{'error': str}`` instead, which the client re-raises.
+"""
+
+import logging
+from abc import ABC, abstractmethod
+from typing import Any
+
+from websockets.sync.server import ServerConnection, serve
+
+# This module + ``protocol`` copy into a benchmark's isolated interpreter as a self-contained unit
+# (importing ``positronic.*`` would run the package's installed-version lookup and fail there).
+# Relative when they land as a package, top-level when copied in flat.
+try:
+    from .protocol import decode, encode
+except ImportError:
+    from protocol import decode, encode
+
+logger = logging.getLogger(__name__)
+
+
+class EnvProtocol(ABC):
+    """A benchmark env behind the three methods the server exposes; the wire contract, positronic-free.
+
+    ``reset`` and ``step`` exchange raw plain-data dicts (numpy arrays + scalars) — the canonical<->raw
+    mapping is the client's ``EnvAdapter``, never the server's. Heavy, per-task construction is the
+    env's own concern (cache it, keyed by the token's structural part); the protocol has no build phase.
+    """
+
+    @abstractmethod
+    def reset(self, token: Any) -> dict[str, Any]:
+        """Construct (cached) + re-randomize from an opaque token; returns ``{'obs', 'meta', 'control_dt'}``.
+
+        ``control_dt`` is the env's control period: the client paces one ``step`` per ``control_dt`` and
+        advances the World's virtual clock by it each step. ``meta`` is whatever the embodiment needs
+        that isn't an observation (e.g. the robot's URDF / joint names).
+        """
+
+    @abstractmethod
+    def step(self, action: dict[str, Any]) -> dict[str, Any]:
+        """Apply a raw action for one control period; returns ``{'obs', 'done', 'control_dt'}``.
+
+        ``control_dt`` is re-reported every step (the wait until the next one), so it can vary.
+        """
+
+    @abstractmethod
+    def close(self) -> None:
+        """Release the env's resources."""
+
+
+class EnvServer:
+    """Serves one ``EnvProtocol`` over a synchronous websocket, lockstep one request at a time.
+
+    The env persists across client connections — its per-task construction cache survives a reconnect;
+    ``shutdown`` releases it.
+    """
+
+    def __init__(self, env: EnvProtocol, host: str, port: int):
+        self._env = env
+        self._host = host
+        self._port = port
+        self._server = None
+
+    def _handle(self, connection: ServerConnection) -> None:
+        for raw in connection:
+            msg = decode(raw)
+            # The wire boundary: a failure handling the command (an env that rejects an action, a sim
+            # blow-up, an unknown command) crosses back as an error frame the client re-raises, rather
+            # than killing the connection.
+            try:
+                match msg['cmd']:
+                    case 'reset':
+                        result = self._env.reset(msg['token'])
+                    case 'step':
+                        result = self._env.step(msg['action'])
+                    case 'close':
+                        connection.send(encode({'ok': True}))
+                        return
+                    case other:
+                        raise ValueError(f'Unknown command: {other!r}')
+            except Exception as e:
+                result = {'error': f'{type(e).__name__}: {e}'}
+            connection.send(encode(result))
+
+    def serve_forever(self) -> None:
+        # A reset token can be a large opaque blob (e.g. an exact-replay scene), so don't cap frame size.
+        with serve(self._handle, self._host, self._port, max_size=None) as server:
+            self._server = server
+            server.serve_forever()
+
+    def shutdown(self) -> None:
+        if self._server is not None:
+            self._server.shutdown()
+        self._env.close()
--- a/positronic/simulator/env_server/tests/mujoco_env.py
+++ b/positronic/simulator/env_server/tests/mujoco_env.py
@@ -1,0 +1,251 @@
+"""The native test fixture: ``MujocoSim``/``stack_cubes`` wrapped behind the env-server protocol.
+
+This is the duplicated, test-only remote copy of the production native path (in-process
+``stack_cubes`` stays the real one), and the reference ``EnvAdapter`` — the simplest one, no
+OSC/axis-angle quirks. ``MujocoEnv`` drives ``MujocoSim.run()`` directly, binding local pipes to its
+ports the way ``World.start`` does, and exposes it as the gym-style ``reset``/``step`` the server
+serves. ``make_mujoco_env`` builds it; ``StackCubesAdapter`` + ``remote_stack_cubes_embodiment`` are
+the client-side mapping + embodiment.
+"""
+
+from collections import deque
+from typing import Any
+
+import numpy as np
+
+import pimm
+import positronic.cfg.simulator
+from pimm.world import LocalQueueEmitter, LocalQueueReceiver, VirtualClock
+from positronic import geom
+from positronic.dataset.serializers import Serializers
+from positronic.drivers.roboarm import command as roboarm_command
+from positronic.eval import ROBOT_STATIC_META, Command, Embodiment, Eval, Observation, Task
+from positronic.simulator.env_server.adapter import EnvAdapter
+from positronic.simulator.env_server.proxy import RemoteEnvControlSystem
+from positronic.simulator.env_server.server import EnvProtocol
+from positronic.simulator.mujoco.sim import MujocoFrankaState, MujocoSim
+from positronic.utils import package_assets_path
+
+_ROTMAT = geom.Rotation.Representation.ROTATION_MATRIX
+
+# Logical observation name -> the model camera the sim renders.
+CAMERAS = {'image.agentview': 'agentview'}
+
+
+class _NeverStop(pimm.SignalReceiver):
+    """A ``should_stop`` that never fires — ``MujocoEnv`` drives the sim loop tick by tick instead."""
+
+    def read(self) -> pimm.Message:
+        return pimm.Message(False, 0, False)
+
+
+class MujocoEnv(EnvProtocol):
+    """A ``MujocoSim`` behind the gym-style env protocol, driving its ``run()`` loop one tick at a time.
+
+    Binding local pipes to the sim's ports (the wiring ``World.start`` performs) lets ``step`` inject a
+    raw action and pump the fused per-tick loop, then read the post-step signals back. The control
+    period is the sim's own physics ``timestep`` — the env has no other dt — reported from ``reset``.
+    """
+
+    def __init__(self, sim: MujocoSim, camera_names: list[str]):
+        self._sim = sim
+        self._clock = VirtualClock()
+        self._gen = None
+        self._cmd_emit = self._bind_input(sim.commands)
+        self._grip_emit = self._bind_input(sim.target_grip)
+        self._state_recv = self._bind_output(sim.state)
+        self._grip_recv = self._bind_output(sim.grip)
+        self._meta_recv = self._bind_output(sim.robot_meta)
+        self._sim_state_recv = self._bind_output(sim.sim_state)
+        self._camera_recvs = {name: self._bind_output(sim.cameras[name]) for name in camera_names}
+
+    def _bind_input(self, receiver: pimm.ControlSystemReceiver) -> LocalQueueEmitter:
+        queue: deque = deque(maxlen=1)
+        receiver._bind(LocalQueueReceiver(queue))
+        return LocalQueueEmitter(queue, self._clock)
+
+    def _bind_output(self, emitter: pimm.ControlSystemEmitter) -> LocalQueueReceiver:
+        queue: deque = deque(maxlen=1)
+        emitter._bind(LocalQueueEmitter(queue, self._clock))
+        return LocalQueueReceiver(queue)
+
+    @property
+    def _timestep(self) -> float:
+        return self._sim.model.opt.timestep
+
+    def _advance(self, duration: float) -> None:
+        target_ns = self._clock.now_ns() + max(1, round(duration * 1e9))
+        while self._clock.now_ns() < target_ns:
+            command = next(self._gen)
+            step_ns = max(1, round(command.seconds * 1e9)) if isinstance(command, pimm.Sleep) else 1
+            self._clock.advance_to_ns(self._clock.now_ns() + step_ns)
+
+    def _read_obs(self) -> dict[str, Any]:
+        state = self._state_recv.read().data
+        ee_pose = state.ee_pose
+        # ``q``/``dq`` already copy out of the sim's reused state buffer; ``ee_pose`` translation is a
+        # live view into it, so copy here too — otherwise the next ``step`` overwrites this obs.
+        return {
+            'q': state.q,
+            'dq': state.dq,
+            'ee_pos': np.array(ee_pose.translation),
+            'ee_quat': np.array(ee_pose.rotation.as_quat),
+            'status': int(state.status.value),
+            'grip': float(self._grip_recv.read().data),
+            'cameras': {name: recv.read().data.array.copy() for name, recv in self._camera_recvs.items()},
+            'sim_state': dict(self._sim_state_recv.read().data),
+        }
+
+    def reset(self, token: Any) -> dict[str, Any]:
+        # Close the prior generator before rebuilding the scene, so its ``run`` cleanup tears down the old
+        # renderer before ``reset`` creates the new one. ``reset`` loads the scene and arms frame-0; pump the
+        # run loop twice — its setup + first control-period sleep, then the reset turn that publishes frame-0 —
+        # and return that frame without stepping. The first ``step`` advances the generator with the first
+        # action (Gym-style).
+        if self._gen is not None:
+            self._gen.close()
+        self._sim.reset(token)
+        self._gen = self._sim.run(_NeverStop(), self._clock)
+        next(self._gen)  # loop setup + first control-period sleep
+        next(self._gen)  # the reset turn: publishes frame-0 (no step)
+        return {'obs': self._read_obs(), 'meta': dict(self._meta_recv.read().data), 'control_dt': self._timestep}
+
+    def step(self, action: dict[str, Any]) -> dict[str, Any]:
+        assert self._gen is not None, 'step() called before reset()'  # real Gym envs reject step-before-reset
+        unknown = action.keys() - {'reset', 'recover', 'joints', 'pose', 'grip'}
+        if unknown:
+            raise ValueError(f'MujocoEnv got unsupported action keys: {sorted(unknown)}')
+        if 'reset' in action:
+            self._cmd_emit.emit(roboarm_command.Reset())  # home the arm
+        elif 'recover' in action:
+            self._cmd_emit.emit(roboarm_command.Recover())  # clear the error in place
+        elif 'joints' in action:
+            self._cmd_emit.emit(roboarm_command.JointPosition(np.asarray(action['joints'], dtype=np.float64)))
+        elif 'pose' in action:
+            self._cmd_emit.emit(
+                roboarm_command.CartesianPosition(geom.Transform3D.from_vector(action['pose'], _ROTMAT))
+            )
+        if 'grip' in action:
+            self._grip_emit.emit(float(action['grip']))
+        self._advance(self._timestep)
+        return {'obs': self._read_obs(), 'done': False, 'control_dt': self._timestep}
+
+    def close(self) -> None:
+        if self._gen is not None:
+            self._gen.close()
+            self._gen = None
+
+
+def make_mujoco_env(camera_names: list[str]) -> MujocoEnv:
+    """Build the ``stack_cubes`` ``MujocoSim`` and wrap it for the protocol."""
+    sim = MujocoSim(
+        package_assets_path('assets/mujoco/franka_table.xml'),
+        positronic.cfg.simulator.stack_cubes_loaders(),
+        camera_resolution=(64, 64),
+        camera_fps=None,
+    )
+    return MujocoEnv(sim, camera_names)
+
+
+class StackCubesAdapter(EnvAdapter):
+    """The reference adapter: canonical Franka commands/observations <-> the MujocoSim raw payloads."""
+
+    def __init__(self, camera_dict: dict[str, str]):
+        self._camera_dict = camera_dict  # logical observation name -> the env's model camera name
+        self._players: dict[str, roboarm_command.TrajectoryPlayer] = {}
+        self._held: dict[str, Any] = {}  # last sampled waypoint per channel — re-sent until it changes
+
+    def reset_token(self, seed: int | None) -> Any:
+        self._players = {}
+        self._held = {}
+        return seed
+
+    def action(self, commands: dict[str, pimm.Message], now_ns: int) -> dict[str, Any]:
+        for name, msg in commands.items():
+            player = self._players.setdefault(name, roboarm_command.TrajectoryPlayer())
+            if msg.updated and msg.data is not None:
+                player.set(msg.data)
+                if not msg.data:  # an empty trajectory cancels: stop replaying the held waypoint
+                    self._held.pop(name, None)
+            for value in player.advance(now_ns):
+                self._held[name] = value
+        # Position setpoints are held and re-sent every control tick — absolute-mode by design, so a
+        # relative delta would integrate repeatedly. Reset/Recover are one-shot events: emitted once,
+        # then dropped so they neither re-fire nor leave a stale setpoint holding behind them.
+        action: dict[str, Any] = {}
+        match self._held.get('robot_command'):
+            case None:
+                pass
+            case roboarm_command.JointPosition(positions):
+                action['joints'] = np.asarray(positions, dtype=np.float64)
+            case roboarm_command.CartesianPosition(pose):
+                action['pose'] = pose.as_vector(_ROTMAT)
+            case roboarm_command.Reset():
+                action['reset'] = True
+                self._held.pop('robot_command')
+            case roboarm_command.Recover():
+                action['recover'] = True
+                self._held.pop('robot_command')
+            case other:
+                raise ValueError(f'StackCubesAdapter cannot map robot_command {type(other).__name__}')
+        grip = self._held.get('target_grip')
+        if grip is not None:
+            action['grip'] = float(grip)
+        return action
+
+    def observations(self, raw_obs: dict[str, Any]) -> dict[str, Any]:
+        state = MujocoFrankaState()
+        ee_pose = geom.Transform3D(raw_obs['ee_pos'], geom.Rotation.from_quat(raw_obs['ee_quat']))
+        state.encode(raw_obs['q'], raw_obs['dq'], ee_pose)
+        state.array[14 + 7] = float(raw_obs['status'])
+        obs: dict[str, Any] = {'robot_state': state, 'grip': float(raw_obs['grip'])}
+        for logical, model_name in self._camera_dict.items():
+            frame = raw_obs['cameras'][model_name]
+            adapter = pimm.shared_memory.NumpySMAdapter(shape=frame.shape, dtype=frame.dtype)
+            adapter.array[:] = frame
+            obs[logical] = adapter
+        return obs
+
+    def privileged(self, raw_obs: dict[str, Any]) -> dict[str, Any]:
+        return {'sim_state': raw_obs['sim_state']}
+
+    def terminal(self, result: dict[str, Any]) -> dict[str, Any] | None:
+        return None  # native stack_cubes scores downstream — it reports no live terminal
+
+
+def remote_stack_cubes_embodiment(proxy: RemoteEnvControlSystem, camera_dict: dict[str, str]) -> Embodiment:
+    """The remote twin of ``mujoco_franka``: the same canonical contract over the proxy's ports."""
+    observations = {
+        'robot_state': Observation(proxy.observations['robot_state'], Serializers.robot_state),
+        'grip': Observation(proxy.observations['grip'], None),
+        **{logical: Observation(proxy.observations[logical], Serializers.camera_images) for logical in camera_dict},
+    }
+    commands = {
+        'robot_command': Command(proxy.commands['robot_command'], roboarm_command.Reset(), Serializers.robot_command),
+        'target_grip': Command(proxy.commands['target_grip'], 0.0, None),
+    }
+    return Embodiment(
+        descriptor='remote.mujoco.franka',
+        observations=observations,
+        commands=commands,
+        static_meta=dict(ROBOT_STATIC_META),
+        meta_source=proxy.robot_meta,
+        control_systems=(proxy,),
+        simulated=True,
+    )
+
+
+def remote_stack_cubes_eval(host: str, port: int, *, camera_dict: dict[str, str]) -> Eval:
+    """Build the remote ``stack_cubes`` eval (embodiment + task) wired to a running env server."""
+    proxy = RemoteEnvControlSystem(StackCubesAdapter(camera_dict), host, port)
+    embodiment = remote_stack_cubes_embodiment(proxy, camera_dict)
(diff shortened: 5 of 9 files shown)


positronic/simulator/env_server/client.py
```diff
@@ -0,0 +1,61 @@
...
+        *,
+        open_timeout: float = 10.0,
+        connect_deadline: float = 60.0,
+    ):
+        uri = f'ws://{host}:{port}/'
+        deadline = time.monotonic() + connect_deadline
+        backoff = 0.5
+        while True:
+            try:
+                self._ws = connect(uri, open_timeout=open_timeout)
```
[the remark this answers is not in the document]
vertix: Fixed in dc84f17: pass `max_size=None` to the client `connect` (it `recv`s the obs responses) and to the server `serve` (a reset token can be a large opaque replay blob). The single 64×64-frame test didn't exercise the 1 MiB default; real multi-camera + full `sim_state` payloads would have. Good catch.

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,72 @@
...
+    def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[pimm.Sleep]:
+        self._players = {name: TrajectoryPlayer() for name in self.commands}
+        try:
+            while not should_stop.value:
+                for name, receiver in self.commands.items():
+                    msg = receiver.read()
+                    if msg.updated and msg.data is not None:
+                        self._players[name].set(msg.data)
+                    for value in self._players[name].advance(clock.now_ns()):
+                        self._held[name] = value
```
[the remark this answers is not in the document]
vertix: Fixed in bb9d00f: an empty `[]` trajectory now drops the channel's held waypoint, so a canceled channel stops replaying its last action — the env then holds via its own mechanism, matching native `MujocoSim`, which applies nothing on a cleared player and relies on sticky actuator ctrl. (The remote path is forced to absolute-mode commands by the hold-last-commanded invariant, so the `JointDelta` case is out of scope — but clearing on cancel is the correct, native-matching behavior regardless.)

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,72 @@
...
+                result = self._conn.step(self._adapter.action(self._held))
```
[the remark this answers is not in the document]
vertix: Fixed in bb9d00f: a `_terminal` flag stops the proxy stepping once the env reports `done`, until the next `reset` — so a step never reaches a terminal Gym env that would reject it. The `MujocoEnv` fixture now asserts `not done` on entry to `step`, so the existing `done`-termination e2e guards against a regression.

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,79 @@
...
+    def _step_env(self, clock: pimm.Clock) -> None:
+        for name, receiver in self.commands.items():
+            msg = receiver.read()
+            if msg.updated and msg.data is not None:
+                self._players[name].set(msg.data)
+                if not msg.data:  # an empty trajectory cancels the channel: stop replaying the held waypoint
+                    self._held.pop(name, None)
+            for value in self._players[name].advance(clock.now_ns()):
+                self._held[name] = value
+        result = self._conn.step(self._adapter.action(self._held))
```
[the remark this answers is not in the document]
vertix: Fixed in 1020b73: the proxy now builds the commands dict from every channel in `self.commands` with `None` defaults, overlaying held waypoints, before calling `action` — so the documented "every channel present, `None` before its first command" contract holds on the initial tick and after a cancel. (The reference adapter used `.get`, which masked it.)

positronic/simulator/env_server/server.py
```diff
@@ -0,0 +1,102 @@
...
+"""
+
+import logging
+from abc import ABC, abstractmethod
+from collections.abc import Callable
+from typing import Any
+
+from websockets.sync.server import ServerConnection, serve
+
+from positronic.simulator.env_server.protocol import decode, encode
```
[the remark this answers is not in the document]
vertix: Fixed in 8856327: `server.py` and `client.py` now import the codec relatively (`from .protocol import ...`), so `server.py` + `protocol.py` copy into a benchmark's isolated interpreter as a self-contained unit — an absolute `positronic.*` import would run `positronic/__init__.py`'s `importlib.metadata.version('positronic')` and fail where positronic isn't installed. Verified: importing the copied pair pulls in zero `positronic` modules, only websockets/msgpack/numpy.

positronic/simulator/env_server/server.py
```diff
@@ -0,0 +1,104 @@
...
+import logging
+from abc import ABC, abstractmethod
+from collections.abc import Callable
+from typing import Any
+
+from websockets.sync.server import ServerConnection, serve
+
+# Relative so this module + ``protocol`` copy into a benchmark's isolated interpreter as a
+# self-contained unit — importing ``positronic.*`` would run the package's installed-version lookup.
+from .protocol import decode, encode
```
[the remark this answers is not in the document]
vertix: Fixed in c06f3c6: `server.py` now tries the relative codec import and falls back to a top-level one, so it loads whether the two files are vendored as a package or copied in flat as top-level modules. Verified both: `EnvServer` imports in each case pulling in zero `positronic` modules.

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,81 @@
...
+        self.commands: pimm.ReceiverDict = pimm.ReceiverDict(self, default=None)
+        self.observations: pimm.EmitterDict = pimm.EmitterDict(self)
+        self.robot_meta: pimm.SignalEmitter = pimm.ControlSystemEmitter(self)
+        self.done: pimm.SignalEmitter = pimm.ControlSystemEmitter(self)
+
+        # The last sampled waypoint per channel: a commanded channel holds its value, so the env
+        # re-applies it every step until a new waypoint arrives or the channel is canceled.
+        self._held: dict[str, Any] = {}
+        self._players: dict[str, TrajectoryPlayer] = {}
+        # Set once the env reports terminal: stepping stops (Gym envs reject step-after-done) until reset.
```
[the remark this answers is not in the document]
vertix: Fixed in 2d30b27: replaced `_terminal` with a single `_active` flag that gates stepping to a live trial — false until the first `reset`, false again after the env's `done`. So an attended run (device loop up before any RUN→reset) no longer steps a never-reset env, and the post-done window no longer steps either. The fixture now also asserts `step()` before `reset()`.

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,81 @@
...
+            msg = receiver.read()
+            if msg.updated and msg.data is not None:
+                self._players[name].set(msg.data)
+                if not msg.data:  # an empty trajectory cancels the channel: stop replaying the held waypoint
+                    self._held.pop(name, None)
+            for value in self._players[name].advance(clock.now_ns()):
+                self._held[name] = value
+        # Every channel present, ``None`` until it holds a waypoint — the ``EnvAdapter.action`` contract.
+        commands = {name: self._held.get(name) for name in self.commands}
+        result = self._conn.step(self._adapter.action(commands))
```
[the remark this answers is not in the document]
vertix: Fixed in 2d30b27: dropped `JointDelta` from the reference adapter and `MujocoEnv`. You're right — the proxy re-sends the held waypoint every control tick, which only composes with absolute commands; a relative delta integrates repeatedly. The remote boundary is absolute-mode OSC by design (`JointPosition`/`CartesianPosition`), so mapping a relative command there was the contradicting pathway, now removed. Added a comment stating the constraint.

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,82 @@
...
+        self._players: dict[str, TrajectoryPlayer] = {}
+        # Step only while a trial is live: not before the first reset, not after the env's done —
+        # Gym envs reject step-before-reset and step-after-done.
+        self._active = False
+
+    def reset(self, seed: int | None = None) -> None:
+        """Re-randomize the env from ``seed`` — the Task's ``RUN``-time reset hook — then emit the
+        fresh scene meta and the post-reset observations."""
+        self._held = {}
+        self._active = True
```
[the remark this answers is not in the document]
vertix: Fixed in 8ac9e91: the proxy now skips exactly the first post-reset tick — the one that lands in the Harness's `reset`→`Yield`→`START` window — so it doesn't step (or emit `done`) before the episode is recording. `reset` already emits the post-reset observations, so the first inference still sees post-reset state; the env's first real advance is then driven by the first policy command, after START.

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,89 @@
...
+            for value in self._players[name].advance(clock.now_ns()):
+                self._held[name] = value
+        # Every channel present, ``None`` until it holds a waypoint — the ``EnvAdapter.action`` contract.
+        commands = {name: self._held.get(name) for name in self.commands}
+        result = self._conn.step(self._adapter.action(commands))
+        self._emit_observations(result['obs'])
+        payload = self._adapter.terminal(result)
+        if payload is not None:
+            self.done.emit(payload)
+            self._active = False
```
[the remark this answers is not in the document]
vertix: Valid — a real cross-trial leak. If a trial ends by timeout (not env-`done`), `_active` stays set (only env-`done` clears it), so the proxy keeps stepping; a `done` the env returns during that overrun — after the Harness already fired the timeout FINISH (`_running` now False, so it's never consumed) — stays queued and the next self-driven trial reads it, its later deadline making the old timestamp look in-budget.

The clean fix is in the Harness's 10a done-read (drain `self.done` at RUN, or window the timestamp against trial-start), since the proxy has no trial-end signal to stop on without adding a hook to the Task/harness contract. That touches merged 10a code, beyond this PR's additive scope, so I'm checking with the author on whether to land it here or as a follow-up. Leaving this open.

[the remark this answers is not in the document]
vertix: Fixed in bfaf15c. The drain lives in the harness, not the proxy: the stale `done` is a message already delivered onto the wire, and the proxy holds the emitter end — only the consumer can clear it.

A deadline-ended trial stops reading `done` (the read is gated on `_running`, which the timeout clears), but the env keeps stepping through the gap before the next trial, so a late terminal can sit queued. The next trial now discards any pending `done` at RUN, right before it re-arms — completing the lifecycle: stop listening at end (already there) → flush at RUN (new) → listen again (already there). The flush has to be at RUN rather than trial-end precisely because the env emits during that gap.

positronic/simulator/env_server/protocol.py
```diff
@@ -0,0 +1,37 @@
...
+            raise ValueError(f'Unsupported dtype: {obj.dtype}')
+        return {b'__ndarray__': True, b'data': obj.tobytes(), b'dtype': obj.dtype.str, b'shape': obj.shape}
+    if isinstance(obj, np.generic):
+        return {b'__npgeneric__': True, b'data': obj.item(), b'dtype': obj.dtype.str}
+    return obj
+
+
+def _unpack(obj):
+    if b'__ndarray__' in obj:
+        return np.ndarray(buffer=obj[b'data'], dtype=np.dtype(obj[b'dtype']), shape=obj[b'shape'])
```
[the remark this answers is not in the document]
vertix: Fixed in bd5c629. `_unpack` now backs the array with a `bytearray` instead of the raw msgpack `bytes`, so decoded arrays are writable — the socket path matches the in-process path for envs/adapters that mutate a decoded buffer in place. Verified a round-tripped array is `flags.writeable` and accepts in-place assignment.

positronic/simulator/env_server/server.py
```diff
@@ -0,0 +1,108 @@
+"""The dumb remote env-server: a benchmark env behind ``reset``/``step``/``close`` over websockets.
+
+Positronic-free by contract — it depends only on ``websockets``, plus the wire codec and the
+``EnvProtocol`` an env implements. A benchmark runs this in its own isolated interpreter (where
+positronic can't be installed); the native fixture runs it in-process against ``MujocoSim``. The
+server is lockstep request-response — one client, one outstanding request — so the World's virtual
+clock advances unchanged while a step round-trips.
+
+Protocol (msgpack frames, see ``protocol``):
+  client -> ``{'cmd': 'make', 'spec': {...}}``      server -> ``{'status': 'ready', 'control_dt', 'meta'}``
```
vertix: Please also document how `control_dt` and `meta` are used, i.e. what do they influence?

positronic/simulator/env_server/tests/mujoco_env.py
```diff
@@ -0,0 +1,234 @@
...
+    Binding local pipes to the sim's ports (the wiring ``World.start`` performs) lets ``step`` inject a
+    raw action and pump the fused per-tick loop for one ``control_dt``, then read the post-step
+    signals back. ``done_after_steps`` fires a synthetic ``done`` to exercise the early-stop path.
+    """
+
+    def __init__(
+        self, sim: MujocoSim, camera_names: list[str], *, control_dt: float, done_after_steps: int | None = None
+    ):
+        self._sim = sim
+        self._control_dt = control_dt
```
vertix: As far as I know, MujocoSim already have the predefined dt, and anything but it would contradict

positronic/simulator/env_server/tests/mujoco_env.py
```diff
@@ -0,0 +1,234 @@
...
+        self._done_after = done_after_steps
```
vertix: This sim does not expose it now, no need to intro this

positronic/simulator/env_server/tests/mujoco_env.py
```diff
@@ -0,0 +1,234 @@
...
+        emitter._bind(LocalQueueEmitter(queue, self._clock))
+        return LocalQueueReceiver(queue)
+
+    @property
+    def control_dt(self) -> float:
+        return self._control_dt
+
+    @property
+    def meta(self) -> dict[str, Any]:
+        return {'initial_ctrl': [float(x) for x in self._sim.initial_ctrl]}
```
vertix: This should be the meta that the sim exposes in our existing sim.stack_cubes task, not smth that we invented here

positronic/simulator/env_server/server.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class EnvServer:
+    """Serves one ``EnvProtocol`` per connection over a synchronous websocket.
+
+    ``env_factory`` builds the env from the client's ``make`` spec, so a single server binary serves
+    whichever scene/task the eval config selects.
+    """
+
+    def __init__(self, env_factory: Callable[[dict[str, Any]], EnvProtocol], host: str, port: int):
```
vertix: Please explain why we accept factory instead of accepting just the protocol itself. I was thinking that specification is something that we get during `reset`.

positronic/simulator/env_server/server.py
vertix: Documented in 351a71f. Both now ride the `reset` response (the protocol lost its handshake — see the factory thread): `control_dt` is the env's control period — the client paces one `step` per `control_dt` and advances the World's virtual clock by it each step; `meta` is whatever the embodiment needs that isn't an observation (for stack_cubes, the sim's `robot_meta`). The `EnvProtocol.reset` docstring spells this out.

positronic/simulator/env_server/tests/mujoco_env.py
vertix: Fixed in 351a71f. `control_dt` is now `sim.model.opt.timestep` — the sim's only dt — returned from `reset` rather than a spec constant. One `step` now advances one physics tick, matching the in-process sim's granularity.

positronic/simulator/env_server/tests/mujoco_env.py
vertix: Fixed in 351a71f. Dropped `done_after_steps`; `MujocoEnv`/`StackCubesAdapter` are the faithful wrapper now (`step` always reports `done=False`, `terminal` returns `None` — stack_cubes scores downstream). The proxy's terminal→`done` path is tested in isolation by a degenerate `_CountdownEnv` in the test file, so no fabricated terminal touches the real wrapper.

positronic/simulator/env_server/tests/mujoco_env.py
vertix: Fixed in 351a71f. The handshake meta and the invented `initial_ctrl` are gone — `reset` returns the sim's own `robot_meta` over the wire, and the remote arm homes via `Reset()` like the real droid. Nobody needs `initial_ctrl` anymore.

positronic/simulator/env_server/server.py
vertix: Redesigned in 351a71f per our discussion — you were right, the factory + `make` phase didn't earn its place. The protocol has no build phase now: `EnvServer` takes an `EnvProtocol` instance, and `reset(token)` constructs (cache heavy things) + re-randomizes, returning `control_dt` + `meta` + `obs`. Per-task `control_dt` is free (rides each reset), and the structural-vs-episodic split moves into the env's own cache (keyed by the token's structural part) instead of onto the wire.

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def _emit_observations(self, raw_obs: dict[str, Any]) -> None:
+        for name, value in self._adapter.observations(raw_obs).items():
+            self.observations[name].emit(value)
+
+    def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[pimm.Sleep]:
+        self._players = {name: TrajectoryPlayer() for name in self.commands}
+        try:
+            while not should_stop.value:
+                if self.control_dt is None:
```
vertix: control_dt should NEVER be None. Again, defensive programming

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,96 @@
...
+        self.control_dt: float | None = None
+
+        self.commands: pimm.ReceiverDict = pimm.ReceiverDict(self, default=None)
+        self.observations: pimm.EmitterDict = pimm.EmitterDict(self)
+        self.robot_meta: pimm.SignalEmitter = pimm.ControlSystemEmitter(self)
+        self.done: pimm.SignalEmitter = pimm.ControlSystemEmitter(self)
+
+        # The last sampled waypoint per channel: a commanded channel holds its value, so the env
+        # re-applies it every step until a new waypoint arrives or the channel is canceled.
+        self._held: dict[str, Any] = {}
```
vertix: This is not the business of the ControlSystem, each adapter should know istself what to do when there are no new actions.
It should also do the trajectory playing, etc.
The control system is just very dumb and translates between signals and adapter, while adapter adapts it.

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,96 @@
...
+        self._players: dict[str, TrajectoryPlayer] = {}
+        # Step only while a trial is live: not before the first reset, not after the env's done —
+        # Gym envs reject step-before-reset and step-after-done.
+        self._active = False
+        # Skip the post-reset tick that lands in the Harness's reset->START yield, so the env isn't
```
vertix: I don't understand, explain simpler

positronic/simulator/env_server/proxy.py
vertix: Fixed in cabc37f — gone. The loop gates on `_active` (is a trial live?) instead of on `control_dt is None`. Before the first reset the proxy isn't in a trial, so it just yields `Yield()` — pimm's `interleave` carries a yielding loop to the next scheduled tick (no spin, no time advance), so the proxy never touches `control_dt` before it's known. `control_dt` is read only while a trial is live, where reset has always set it. No nullable, no defensive check.

positronic/simulator/env_server/proxy.py
vertix: Done in cabc37f. `RemoteEnvControlSystem` is a dumb translator now: each tick it reads the command receivers, hands the messages to `adapter.action(commands, now_ns)`, ships the raw action it returns, and re-emits whatever `adapter.observations` maps back. The `TrajectoryPlayer`s, the hold-last-waypoint, and the canonical→raw mapping all moved into the adapter — `StackCubesAdapter` owns them now, so each adapter decides for itself what to do when there's no new command. The control system keeps only the pimm-port plumbing and the trial lifecycle.

positronic/simulator/env_server/proxy.py
vertix: Plain version: a trial goes reset → (one tick) → start recording → run. The proxy steps the env every tick. Without the skip it would step the env during that one-tick gap — before recording starts — and a fast env could even report `done` there, ending the trial before it records anything. `_skip_step` makes the proxy sit out exactly that one post-reset tick. Reworded the comment in cabc37f to say this directly.

It's the one place the proxy has to know about the Harness's record boundary — if you see a cleaner way to express that coupling, I'm open to it.

positronic/simulator/env_server/server.py
```diff
@@ -0,0 +1,98 @@
...
+    """A benchmark env behind the three methods the server exposes; the wire contract, positronic-free.
+
+    ``reset`` and ``step`` exchange raw plain-data dicts (numpy arrays + scalars) — the canonical<->raw
+    mapping is the client's ``EnvAdapter``, never the server's. Heavy, per-task construction is the
+    env's own concern (cache it, keyed by the token's structural part); the protocol has no build phase.
+    """
+
+    @abstractmethod
+    def reset(self, token: Any) -> dict[str, Any]:
+        """Construct (cached) + re-randomize from an opaque token; returns ``{'obs', 'meta', 'control_dt'}``.
```
vertix, line 47: I think we should return control_dt as a part of observation. Who knows, may be some weird sim in the future will want to use a non-uniform dt, I just think this API is more general without overcomplication.
The only question is how this should be modelled in the control system? Should we ouput observations before the tick? Or after the tick? Likely after the tick.

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,73 @@
...
+        self.observations: pimm.EmitterDict = pimm.EmitterDict(self)
+        self.robot_meta: pimm.SignalEmitter = pimm.ControlSystemEmitter(self)
+        self.done: pimm.SignalEmitter = pimm.ControlSystemEmitter(self)
+
+        # A trial is live between reset and the env's done. The proxy steps only then — not before the
+        # first reset (Gym envs reject step-before-reset), not after done.
+        self._active = False
+        # The env reports its control period at every reset, so it is set whenever a trial is live.
+        self._control_dt = 0.0
+        # After reset the Harness has a one-tick gap before it starts recording. Skip the env step on
```
vertix: It looks like a hack that has to address another hack. Instead of building them on each other shoulders, is there a right way to fix it in Harness?
Can you explain in simple terms what happens there, so that we find the solution?

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,73 @@
...
+from positronic.simulator.env_server.client import EnvConnection
+
+
+class RemoteEnvControlSystem(pimm.ControlSystem):
+    def __init__(self, adapter: EnvAdapter, host: str, port: int):
+        self._adapter = adapter
+        self._conn = EnvConnection(host, port)
+
+        self.commands: pimm.ReceiverDict = pimm.ReceiverDict(self, default=None)
+        self.observations: pimm.EmitterDict = pimm.EmitterDict(self)
```
vertix, line 30: In tasks we have both observations and priveleged data. We should support privileged data here (and in adapter)

positronic/simulator/env_server/server.py
vertix: Fixed in 8cf0f83. `step` now returns `control_dt` alongside `obs`/`done` (and `reset` keeps returning it), so it rides every observation and a benchmark can vary it per step. The proxy stores the dt from the latest result and sleeps that.

Modelling: "after the tick", as you guessed. The proxy steps, emits the post-step obs — which carries the dt until the *next* obs — then sleeps that dt. So the dt travelling with an observation is the wait until the one after it: `reset`'s obs carries the dt for the first step, each `step`'s obs carries the dt for the following step.

positronic/simulator/env_server/proxy.py
vertix: Fixed in 8cf0f83. Added `EnvAdapter.privileged(raw_obs)` next to `observations(raw_obs)`, plus a `proxy.privileged` EmitterDict. The adapter routes ground-truth (full `sim_state`) through `privileged` and policy-facing signals through `observations`; the proxy emits each into its own dict, and the Task sources `privileged={'sim_state': Observation(proxy.privileged['sim_state'], None)}`.

This is parity with the in-process path: there the sim already exposes `sim.sim_state` as a dedicated emitter (`cfg/eval/sim/positronic.py`), separate from its observation emitters. The remote adapter was the odd one out — it lumped `sim_state` into `observations()`. Now the split is enforced at the adapter boundary, so privileged data can't reach the policy.

positronic/simulator/env_server/proxy.py
vertix: Plain version of what happens at trial start. The Harness does three things in order:

1. reset the scene
2. yield one World tick — so devices publish their post-reset state before the first inference
3. start recording

The proxy is a free-running control system: every tick it advances the env by one `control_dt`. So on that step-2 tick it would advance the env a full step *before* recording has started — and if the env reports `done` there, the `done` sits queued and instantly ends the fresh trial. `_skip_step` makes the proxy sit out exactly that one tick.

You're right it's a hack compensating for another hack. The "other hack": the in-process `MujocoSim` does the exact same thing — `run()` calls `self.step()` on every tick, including that gap tick — but it gets away with it because `stack_cubes` scores `done` downstream (never fires mid-trial), so the one-step drift is invisible. The remote proxy can't get away with it because its `done` rides the step.

So the right fix is in the Harness/contract, not the proxy. Two options, and I'd like your steer before touching it:

- **Cleanest (my preference):** make "a reset publishes frame-0 without stepping" the env-loop contract for *both* embodiments, and have the Harness start recording at that frame. Then step-2's special Yield and the proxy's `skip_step` both disappear — the proxy just steps every `control_dt` while active, with zero knowledge of Harness timing, and the in-process sim stops silently drifting a step too. Cost: it touches `MujocoSim.run()` and the Harness RUN ordering, i.e. shared behavior beyond this PR's proxy.
- **Lighter, Harness-only:** drain `done` *after* the post-reset Yield instead of before it. That kills the cross-trial leak with no sim change and removes `skip_step`, but the env still physics-drifts one step in the gap — a smaller fix for a smaller part of the problem.

Which way do you want to go?

positronic/simulator/env_server/tests/mujoco_env.py
```diff
@@ -174,14 +174,17 @@ def observations(self, raw_obs: dict[str, Any]) -> dict[str, Any]:
...
+        obs: dict[str, Any] = {'robot_state': state, 'grip': float(raw_obs['grip'])}
         for logical, model_name in self._camera_dict.items():
             frame = raw_obs['cameras'][model_name]
             adapter = pimm.shared_memory.NumpySMAdapter(shape=frame.shape, dtype=frame.dtype)
             adapter.array[:] = frame
             obs[logical] = adapter
         return obs
 
+    def privileged(self, raw_obs: dict[str, Any]) -> dict[str, Any]:
+        return {'sim_state': raw_obs['sim_state']}  # full physics state — the ground-truth the in-process sim exposes
```
vertix: no need for this comment, the code says it all

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,79 @@
...
+                    yield pimm.Sleep(self._control_dt)
+                else:
+                    yield pimm.Yield()  # no live trial — ride along to the next tick without stepping
+        finally:
+            self._conn.close()
+
+    def _step_env(self, clock: pimm.Clock) -> None:
+        commands = {name: receiver.read() for name, receiver in self.commands.items()}
+        result = self._conn.step(self._adapter.action(commands, clock.now_ns()))
+        self._control_dt = result['control_dt']
```
vertix: It seems that this method can just return control_dt, in that case we don't need to have control_dt as class attribute

positronic/simulator/env_server/proxy.py
vertix: I want cleanest

vertix: Done in fd64fae — the cleanest fix. `skip_step` is gone. What changed:

- **Harness** (`harness.py`): `START` + `_running` now fire *before* the post-reset tick (was after). So the env publishes its post-reset frame-0 *into* the live recording, and its first step lands inside the trial. Real embodiments (`reset=None`) are untouched — they never had this tick.
- **Proxy** (`proxy.py`): no `skip_step`, no harness-timing knowledge. The loop is `publish current frame → Sleep(its control_dt) → step → next frame`. `reset` produces frame-0; each `step` the next. The one bit of state left, `_needs_step`, is the env's *own* rule ("a freshly reset frame publishes before any step"), not a compensation for the harness.
- **`MujocoSim.run()`**: emits the current instant at the top of the loop, then steps — so the in-process sim records frame-0 too and the silent one-step drift is gone. Both embodiments now behave the same.

Full suite green (585 passed). The golden is unaffected (it drives a fake robot with no `task.reset`).

positronic/simulator/env_server/proxy.py
vertix: Resolved by fd64fae — `skip_step` is removed entirely. The Harness now starts recording before the env steps, so the env publishes frame-0 into the episode and the proxy needs no knowledge of the record boundary. Detail on the "I want cleanest" thread.

positronic/simulator/env_server/proxy.py
vertix: Done in fd64fae. No `_control_dt` attribute anymore. `_step_env` returns the env's frame (`obs` + `control_dt`) and the run loop reads `control_dt` straight off it for the `Sleep`. It returns the whole frame rather than just `control_dt` because the loop also publishes that frame's `obs` — but the redundant attribute you flagged is gone.

positronic/simulator/env_server/tests/mujoco_env.py
vertix: Done in fd64fae — removed. The `EnvAdapter.privileged` docstring already states ground-truth (full sim state) goes there.

positronic/simulator/env_server/tests/mujoco_env.py
```diff
@@ -0,0 +1,227 @@
...
+            for value in player.advance(now_ns):
+                self._held[name] = value
+        # Absolute commands only: the held waypoint re-sends every control tick, so a relative delta
+        # would integrate repeatedly. The remote boundary is absolute-mode by design.
+        action: dict[str, Any] = {}
+        match self._held.get('robot_command'):
+            case roboarm_command.JointPosition(positions):
+                action['joints'] = np.asarray(positions, dtype=np.float64)
+            case roboarm_command.CartesianPosition(pose):
+                action['pose'] = pose.as_vector(_ROTMAT)
```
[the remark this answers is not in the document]
vertix: Real gap, confirmed. `Reset` (the home action, on HOME/FINISH) and `Recover` (the `ErrorRecovery` wrapper, on an IK-failure ERROR) both reach `action()`, the match has no case for them so they're dropped, and `MujocoEnv.step` wouldn't apply them anyway. A remote Franka that fails IK enters ERROR and can never clear it — good catch.

But the suggested fix — adding `Reset`/`Recover` cases to the match — misfires under this adapter's model. It *holds* the last `robot_command` in `self._held` and re-evaluates the match every control tick (the absolute-mode boundary: held setpoints re-send so the remote stays commanded). A one-shot `Reset` cached in `_held` would therefore re-fire every tick and re-reset the sim forever. `Reset`/`Recover` are edge-triggered control-state transitions, not held setpoints — the native sim gets away with it because it applies commands straight from `TrajectoryPlayer.advance()` (one-shot), never from a held cache.

So a correct fix needs one-shot semantics on the client (apply once, don't re-hold) plus `reset`/`recover` handling on the server side of `MujocoEnv.step` — i.e. a decision on how the absolute-mode boundary carries edge-triggered control state, not just held setpoints. That's the same boundary-vs-lifecycle design I'm working through with the maintainer on the `skip_step` thread, so I'm folding it in there rather than landing a fix that re-fires. Leaving this open until that direction is settled.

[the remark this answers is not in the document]
vertix: Fixed in 4ec2c86 — though not the way suggested (adding `Reset`/`Recover` to the held-command match would re-fire the one-shot every tick, as noted above). What landed:

- **Adapter**: `Reset`/`Recover` now map to `{'reset': True}` / `{'recover': True}`, fired **once** then dropped — so they don't re-fire and don't leave a stale setpoint holding. Held setpoints (joint/Cartesian) are unchanged. Any other command type raises (fail loud).
- **`MujocoEnv.step`**: `recover` → clear the arm's error in place; `reset` → home the arm. Unknown action keys raise.
- **`MujocoSim`**: the `Reset` command now homes the arm (`_home`) instead of re-randomizing the scene — re-randomization is `reset(token)`'s job, not the home action's.
- **Protocol**: a new error frame — a command whose handling raises crosses back as `{'error': ...}` the client re-raises, instead of killing the socket. Covers unsupported commands and any server-side blow-up (new test: `test_server_failure_crosses_as_error_frame`).

So a remote Franka that hits an IK error now recovers like the native path. Full suite green (586 passed).

positronic/simulator/mujoco/sim.py
```diff
@@ -167,6 +168,20 @@ def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[p
...
         while not should_stop.value:
+            now = clock.now()
+            if state_due(now):
+                state.encode(self._q, self._dq, self._ee_pose)
+                self.state.emit(state)
+            if grip_due(now):
+                self.grip.emit(self._current_grip())
+            if self.sim_state.num_bound and sim_state_due(now):
+                self.sim_state.emit({f'.{name}': arr for name, arr in self.save_state().items() if arr.size})
+            if adapters and cameras_due(now):
```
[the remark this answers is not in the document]
vertix: Sharp catch — confirmed and fixed in c653f1a. You traced it exactly: the in-process `run()` generator is persistent across trials and was suspended at a *mid-loop* `Sleep`, so a between-trial `reset()` resumed into the step phase (read commands → `step()`) before looping back to emit — making trial 2+'s first frame post-step. Trial 1 was fine (fresh generator), and the remote wrapper was fine (it restarts its generator on reset); only the persistent in-process generator had the gap.

Fix: moved the `Sleep` to the **end** of the loop, after `step()`. Now a reset suspends the loop at the end, so the resume re-enters at the emit phase and publishes the reset's frame-0 — for every trial, matching trial 1 and the remote path. The continuously-running emit sequence is unchanged; only the suspension point moved. Full suite green (586 passed).

positronic/policy/harness.py
```diff
@@ -186,57 +186,68 @@ def _cancel_trajectories(self) -> None:
...
+        # ``inference_latency`` rides the RUN context (and lands in episode meta with it).
+        self._inference_latency = self.context.get('inference_latency', False)
+        self._deadline = None  # set by the run loop for self-driven trials only
+        self._session = self.policy.new_session(self.context)
+        resetting = self._task is not None and self._task.reset is not None
+        if resetting:
+            self._task.reset(self.context.get('eval.seed'))  # re-randomize the scene before recording starts
+        self.ds_command.emit(DsWriterCommand.START())
+        self._running = True
+        if resetting:
```
vertix: Do we really need this if? Should not we always Yield?

positronic/policy/harness.py
```diff
@@ -186,57 +186,68 @@ def _cancel_trajectories(self) -> None:
...
-                        yield pimm.Yield()
-                # ``inference_latency`` rides the RUN context (and lands in episode meta with it).
-                self._inference_latency = self.context.get('inference_latency', False)
-                self._deadline = None  # set by the run loop for self-driven trials only
-                if self._session:
-                    self._session.close()
-                self._session = self.policy.new_session(self.context)
-                self.ds_command.emit(DsWriterCommand.START(self._build_episode_meta(self.context)))
-                self._running = True
+                    yield from self._end_episode(clock)
```
vertix: Wait, this is smth new? I think if we are running, we should ignore run? If we finished, we should ingore stop?

Let me know if this breaks smth, but I think this is more logical

positronic/policy/harness.py
```diff
@@ -186,57 +186,68 @@ def _cancel_trajectories(self) -> None:
...
             case DirectiveType.HOME:
-                if self._running:
-                    self.ds_command.emit(DsWriterCommand.ABORT())
-                if self._session:  # HOME aborts the episode; release the session like FINISH
-                    self._session.close()
-                    self._session = None
-                self._home(clock)
-                yield pimm.Yield()
-                self._running = False
+                yield from self._end_episode(clock, abort=True)
```
vertix, line 259: Given that we abort, I think calling it HOME is the wrong naming? WDYT?

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,74 @@
...
+
+        # A trial is live between reset and the env's done. The proxy publishes/steps only then — not
+        # before the first reset (Gym envs reject step-before-reset), not after done.
+        self._active = False
+        # The latest env frame (``obs`` + ``control_dt``), published on the next tick. ``reset`` produces
+        # frame-0; each ``step`` produces the next — so a freshly reset frame is published before any step.
+        self._frame: dict[str, Any] | None = None
+        self._needs_step = False
+
+    def reset(self, seed: int | None = None) -> None:
```
vertix, line 43: Why do we need this function? Who is calling it?

positronic/simulator/mujoco/sim.py
```diff
@@ -167,6 +168,17 @@ def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[p
         cameras_due = _Cadence(self._camera_fps)
 
         while not should_stop.value:
+            now = clock.now()
+            if state_due(now):
+                state.encode(self._q, self._dq, self._ee_pose)
+                self.state.emit(state)
```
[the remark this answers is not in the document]
vertix: Fixed in 10dfd10: `MujocoSim.run` now emits `state.copy()`. The persistent `state` still tracks the sticky error status across ticks, but the published frame is a by-value snapshot taken before `_apply_command` runs — so a same-tick `Recover`/`Reset`/IK-failure mutates only the next tick's encode, never the already-emitted frame. This matches the remote path, where the obs is serialized to bytes (a natural snapshot).

positronic/policy/harness.py
vertix: Kept the `if` (reverted the always-Yield). The Yield exists to give the env one tick to publish its **post-reset** frame-0 into the now-live recording. With no reset (`task.reset is None`, or no task at all) there's nothing to publish, so an unconditional Yield just adds an empty tick — and that tick perturbs timing: it shifts the golden pipeline's exact frame sequence and pushes `test_done_after_deadline_is_a_timeout`'s `done` from past-deadline into budget. So the guard earns its place; it's the same `resetting` predicate that gates the reset call right above.

positronic/policy/harness.py
vertix: Agreed, done in 639d641. `_handle_directive` now ignores a RUN while a trial is live and a FINISH while idle. HOME stays unguarded — it homes the devices even when idle (the "go to safe state" button), aborting the recording only if one is live. Updated `test_run_while_running_auto_finalizes` -> `test_run_while_running_is_ignored` to assert the new semantics (ep2's RUN is dropped; ep1 finalizes once at shutdown).

positronic/simulator/env_server/proxy.py
vertix: It's the Task's scene-reset hook. The remote eval wires `reset=proxy.reset` into the `Task` (mujoco_env.py:242), and the harness calls `task.reset(seed)` once per episode in `_begin_episode` — so `proxy.reset` re-randomizes the env from the trial's seed and primes frame-0 before recording. The in-process sim wires `reset=sim.reset` the same way (cfg/eval/sim/positronic.py:31). So it's invoked once per episode, not dead — it's the remote twin of the native `sim.reset`.

positronic/policy/harness.py
vertix: Agreed, renamed in 82fccf6: `HOME` -> `ABORT` across the directive enum, classmethod, harness dispatch, GUI reset button, keyboard `r`, and tests. It pairs cleanly now — `FINISH` keeps the recording, `ABORT` discards it; both home the devices. Also dropped the dead `preset` arg (the home action always came from `embodiment.home`, never the preset).

positronic/simulator/env_server/tests/mujoco_env.py
```diff
@@ -0,0 +1,245 @@
...
+        elif 'joints' in action:
+            self._cmd_emit.emit(roboarm_command.JointPosition(np.asarray(action['joints'], dtype=np.float64)))
+        elif 'pose' in action:
+            self._cmd_emit.emit(
+                roboarm_command.CartesianPosition(geom.Transform3D.from_vector(action['pose'], _ROTMAT))
+            )
+        if 'grip' in action:
+            self._grip_emit.emit(float(action['grip']))
+        self._advance(self._timestep)
+        return {'obs': self._read_obs(), 'done': False, 'control_dt': self._timestep}
```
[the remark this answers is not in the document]
vertix: Fixed in ea50e7c. `MujocoSim.run` now applies the tick's command, steps, *then* emits — so every frame is the post-action state, and the env-server fixture's `step()` returns post-step obs (Gym-style) with no change to the fixture itself. To keep frame-0 correct (the invariant c653f1a established), the first tick after a reset skips the step, so the recorded frame-0 is the freshly reset scene rather than one step past it. Native in-process and remote sim are now consistent with real Gym envs.

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,74 @@
...
+        """Re-randomize the env from ``seed`` (the Task's ``RUN`` hook); publish the fresh scene meta and
+        hold frame-0 (control period + post-reset observations) for the run loop to publish next tick."""
+        self._active = True
+        self._needs_step = False
+        self._frame = self._conn.reset(self._adapter.reset_token(seed))
```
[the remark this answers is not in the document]
vertix: Fixed in 181cdf6. `reset()` now drains the command receivers before re-enabling stepping, so a home/`Reset` the harness emitted at the prior episode's end (while the proxy was inactive) can't ride into the new trial — combined with `reset_token` clearing the adapter's held waypoints, the first post-reset step applies an empty action until the policy produces one.

positronic/simulator/env_server/client.py
```diff
@@ -0,0 +1,55 @@
...
+Positronic-free (``websockets`` + the wire codec). ``RemoteEnvControlSystem`` wraps this as a pimm
+control system; tests use it directly to compare a socket rollout against an in-process one.
+"""
+
+import time
+from typing import Any
+
+from websockets.sync.client import connect
+
+from .protocol import decode, encode
```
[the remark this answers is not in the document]
vertix: Declining this one. Unlike `server.py` + `protocol.py`, which are copied flat into a benchmark's isolated interpreter (hence their relative/top-level import fallback), the client always runs inside positronic — `RemoteEnvControlSystem` and the tests import it as `positronic.simulator.env_server.client`, so `.protocol` always has package context. There's no flat-vendored client consumer today, so the fallback would guard a layout that doesn't occur. Happy to add it the day a positronic-free client driver gets vendored standalone.

positronic/simulator/mujoco/sim.py
```diff
@@ -208,7 +220,8 @@ def reset(self, seed: int | None = None):
         ``scene_xml``.
         """
         self._load_scene(seed)
-        self._warmup()
+        self._home()
+        self._just_reset = True  # the next tick emits the reset scene as frame-0 without stepping past it
```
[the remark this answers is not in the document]
vertix: Good catch, fixed in 28e722c. The reset tick now forces every stream (state, grip, sim_state, cameras) regardless of its cadence — `due(now) or self._just_reset` — so the freshly reset scene is emitted complete as frame-0. Without it, a finite-fps camera (e.g. the stack_cubes eval's cameras) whose period wasn't due on the reset tick would leave the new episode's first observation showing the previous scene. `_just_reset` is now cleared after the emit block so it gates both the step-skip and the forced emit.

positronic/policy/harness.py
```diff
@@ -186,57 +186,68 @@ def _cancel_trajectories(self) -> None:
...
+        the env's post-reset frame-0 is the recorded first frame."""
+        self.context = context
+        if self._task is not None:
+            self.context = {**self.context, 'task': self._task.instruction}
+            # A deadline-ended trial stops reading ``done`` while its env keeps running, so a late
+            # terminal can sit queued; discard it here so it can't end this fresh trial.
+            self.done.read()
+        # ``inference_latency`` rides the RUN context (and lands in episode meta with it).
+        self._inference_latency = self.context.get('inference_latency', False)
+        self._deadline = None  # set by the run loop for self-driven trials only
```
vertix: Not only for self-driven trials only. Assisted trials also may have the deadlines, I see no problem

vertix: Agreed, done in 4cebebc. The deadline is now a property of the task, not the driving mode: `_begin_episode` sets `self._deadline = clock.now() + self._task.timeout` whenever there's a task (both the self-driven and the directive/RUN paths call it). So an attended trial with a task honors its budget and can be `done`-terminated, while the operator can still FINISH/ABORT earlier; a task-less attended session keeps no deadline.

No behavior change for existing trials — self-driven get the identical deadline (same instant, no yields between the old two lines), and every attended test is task-less. The constructor assert already allowed `task` set with `trials=None`, so this just fills in the deadline for that case. Updated the three comments that claimed "attended = no deadline".

positronic/policy/harness.py
```diff
@@ -334,21 +349,20 @@ def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[p
...
+            # terminal lands within its budget, or on the budget itself. The deadline is hard, so a terminal
+            # past it is a timeout, not a late success — an in-budget terminal records ``eval.terminated``
+            # True plus its payload, a timeout records only False. Attended episodes carry no deadline and
+            # are ended by the operator's directives, so ``done`` never terminates (or leaks across) them.
             if self._running and self._deadline is not None:
                 done_msg = self.done.read()
-                if done_msg.updated and done_msg.ts <= self._deadline * 1e9:
-                    payload = {**done_msg.data, 'eval.terminated': True}
-                    yield from self._handle_directive(Directive.FINISH(**payload), clock)
+                if done_msg.data and done_msg.ts <= self._deadline * 1e9:
```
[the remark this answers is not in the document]
vertix: Two separate issues here — fixed one, and the other is a deliberate design point I'd like your steer on.

**The `{}` ambiguity — fixed in 81c10f5.** You're right that the `terminal` contract was self-contradictory. The docstring said *"delivery is the terminating edge; an empty `{}` still ends the trial"* — pure edge semantics — but the harness ends on a truthy value (`if done_msg.data`), `test_reset_clears_done_across_trials` asserts *"truthiness, not delivery, is the signal,"* and the proxy uses `{}` as its non-terminal **reset clear**. Worse, `_step_env` emitted (and set `_active = False`) on any non-`None` payload, so an adapter returning `{}` would **halt the proxy while the harness kept running** — a hang. Now the proxy emits `done` only on a non-empty payload, and the `EnvAdapter.terminal` docstring states the truthy-value contract (`None`/`{}` continue; a non-empty payload ends and is recorded).

**The stale latch — by design, but flagging the trade-off.** This is the value-semantics decision (the `done` cell *is* the world's terminal state; `reset` transitions it to non-terminal). The proxy honors it (`reset` emits `{}`); the in-process sim has no `done`. So no shipped config latches, and `test_reset_clears_done_across_trials` enforces the contract.

On the suggested "drain on `_begin_episode`" route: under truthy semantics a harness-side drain can't fix it — a receiver advances `updated` but never clears the cell's *value*, and the check reads the value. So the only truthy-compatible options are (a) the producer clears on reset (current), or (b) the harness adds a `ts` **lower bound** (`trial_start <= done.ts <= deadline`) so a terminal stamped before this trial is rejected regardless of the producer. (b) is clean and stays in value semantics, but it makes `reset`-clears non-load-bearing and reshapes the test's premise. I'd rather not override your reset-clears design unilaterally — want me to add the `ts` lower bound as defense-in-depth, or keep the producer-clears contract as is? Leaving this open for your call.

positronic/policy/harness.py
```diff
@@ -186,57 +186,70 @@ def _cancel_trajectories(self) -> None:
...
+            self.context = {**self.context, 'task': self._task.instruction}
+        # ``inference_latency`` rides the RUN context (and lands in episode meta with it).
+        self._inference_latency = self.context.get('inference_latency', False)
+        # A task's ``timeout`` bounds the trial whether the policy or an operator drives it; a task-less
+        # attended session has no deadline and ends only on a directive.
+        self._deadline = clock.now() + self._task.timeout if self._task is not None else None
+        self._session = self.policy.new_session(self.context)
+        if self._task is not None and self._task.reset is not None:
+            self._task.reset(self.context.get('eval.seed'))
+        self.ds_command.emit(DsWriterCommand.START())
```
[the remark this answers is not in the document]
vertix: Fixed in cdf280c — confirmed and reproduced first (the recorded first frame was the post-step frame, not frame-0).

Root cause exactly as you traced: at the trial-start instant the producer (scheduled before the ds_agent) steps and overwrites the obs cell before the recorder samples it. And the deeper trap you flagged — frame-0 and the first post-step frame both carry ts 0 — means simply reordering the recorder ahead of the producer instead crashes on a non-increasing timestamp.

Fix: both producers (`MujocoSim.run` and `RemoteEnvControlSystem.run`) now sleep one control period before stepping, so the frame-0 that `reset` published synchronously is the recorder's first sample, and the first step lands at `control_dt` (no ts collision). The policy still reads frame-0 synchronously in `_step`, so its first inference is unchanged. Guarded by three tests: `test_policy_first_obs_is_frame0` (policy read), `test_remote_eval_records_frame0` (remote recording), and a frame-0 assertion in `test_sim_emits_commands_and_records_dataset` (native recording).

positronic/simulator/env_server/tests/mujoco_env.py
```diff
@@ -0,0 +1,245 @@
...
+            'ee_pos': np.array(ee_pose.translation),
+            'ee_quat': np.array(ee_pose.rotation.as_quat),
+            'status': int(state.status.value),
+            'grip': float(self._grip_recv.read().data),
+            'cameras': {name: recv.read().data.array.copy() for name, recv in self._camera_recvs.items()},
+            'sim_state': dict(self._sim_state_recv.read().data),
+        }
+
+    def reset(self, token: Any) -> dict[str, Any]:
+        self._sim.reset(token)
```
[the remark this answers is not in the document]
vertix: Fixed in cdf280c. `MujocoEnv.reset` now closes the prior generator before `_sim.reset()`, so the old `run` cleanup tears down the old renderer before the new scene builds its own; and `MujocoSim.run`'s `finally` nulls `self._renderer` after closing (matching `_bind_model`), so the next render recreates it instead of reusing a closed one. Good catch — the `FakeRenderer` was masking it.

positronic/simulator/env_server/tests/mujoco_env.py
```diff
@@ -0,0 +1,245 @@
...
+        if self._gen is not None:
+            self._gen.close()
+        self._gen = self._sim.run(_NeverStop(), self._clock)
+        self._advance(self._timestep)
```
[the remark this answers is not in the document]
vertix: Fixed in cdf280c. `MujocoEnv.reset` no longer steps — it returns the frame-0 `reset` published, and the first `step` advances the generator with the first action (Gym-style). The generator is primed once after creation so each `step` still advances exactly one tick, since the producer loop now sleeps before stepping.

positronic/simulator/mujoco/sim.py
```diff
@@ -180,35 +187,36 @@ def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[p
...
+        first inference reads the reset state directly (no tick consumed) before the producer free-runs.
+        Stale commands queued while idle (e.g. the inter-episode home) are dropped so the first step
+        doesn't apply them to the freshly reset scene.
         """
         self._load_scene(seed)
-        self._warmup()
+        self._home()
+        self._error = False
+        self.commands.read()
+        self.target_grip.read()
```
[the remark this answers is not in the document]
vertix: Fixed in 75d074f: `arm_player`/`grip_player`/`last_grip` were `run()` generator locals, so `reset()` could only drain the receivers, not the playback state — the exact asymmetry you flagged (the remote path already clears this via the adapter's `reset_token`). Promoted them to instance state (`_arm_player`/`_grip_player`/`_last_grip`) and `reset()` now `set([])`s both players and zeroes the held grip alongside draining the receivers, so a residual trajectory or last grip from the prior trial can't replay onto the freshly reset scene regardless of whether the sim has ticked since the cancel.

positronic/policy/harness.py
```diff
@@ -186,57 +186,70 @@ def _cancel_trajectories(self) -> None:
...
+        """
+        if self._running:
+            if abort:
+                self.ds_command.emit(DsWriterCommand.ABORT())
+            else:
+                self._finalize_recording(payload)
+            # Let the recorder commit the STOP/ABORT before the next START (they share ``ds_command`` —
+            # without a tick between, last-value-wins would drop one) and before the home command, so
+            # homing stays out of the recording.
+            yield pimm.Yield()
```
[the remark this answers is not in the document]
vertix: Fixed in 75d074f: the ABORT branch now calls `_cancel_trajectories()` before emitting `ABORT` and yielding, mirroring what `_finalize_recording()` does for FINISH. So the `[]` cancel reaches the drivers during the same `yield` that precedes `_home()`, and a buffered trajectory no longer advances for a scheduler slice on abort. Kept it in the abort branch (rather than hoisting out of `_finalize_recording`) to preserve the FINISH ordering of `_on_complete` before the cancel.

positronic/simulator/env_server/proxy.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[pimm.Command]:
+        try:
+            while not should_stop.value:
+                if self._active:
+                    # Nap one control period before stepping, so the frame-0 ``reset`` published
+                    # synchronously is the recorder's first sample; the first step then lands at
+                    # ``control_dt``, not at the reset instant (which would collide with frame-0's ts).
+                    yield pimm.Sleep(self._frame['control_dt'])
+                    self._frame = self._step_env(clock)
```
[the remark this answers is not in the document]
vertix: Fixed in 3ade67e: added a per-reset epoch guard. `reset()` bumps `self._epoch`; the run loop captures it before each nap and, if `reset()` bumped it mid-nap, `continue`s instead of stepping — so the loop re-naps and the recorder samples the just-published frame-0 before the next step lands.

Confirmed it was a real race, not just analogy: with `_CountdownEnv`'s default `control_dt=0.1` the proxy's nap was long enough that the recorder always read frame-0 first, so a 2-trial test passed even unfixed. Dropping `control_dt` to `0.01` (so the proxy resumes from its nap in the same instant the next `reset()` publishes frame-0) reproduced it — episode 2 recorded `value=[1,…]` (one post-step) instead of frame-0 zeros. `test_remote_eval_records_frame0` now runs two trials at that small `control_dt` and asserts every episode starts at frame-0; it fails without the guard and passes with it.

positronic/simulator/mujoco/sim.py
```diff
@@ -155,60 +162,65 @@ def __init__(
...
         state_due = _Cadence(self._state_fps)
         grip_due = _Cadence(self._grip_fps)
         sim_state_due = _Cadence(self._sim_state_fps)
         cameras_due = _Cadence(self._camera_fps)
 
         while not should_stop.value:
+            # Nap a control period before stepping, so the frame-0 ``reset`` published synchronously is the
+            # recorder's first sample and the first step lands at ``timestep`` (not the reset instant, whose
+            # ts would collide with frame-0's).
+            yield pimm.Sleep(self.model.opt.timestep)
```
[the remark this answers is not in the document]
vertix: Fixed in 3ade67e with the same per-reset epoch guard as the proxy: `reset()` bumps `self._epoch`, and the run loop captures it before the nap and `continue`s (re-naps) instead of stepping if a reset landed mid-nap — so the first post-step frame can't overwrite the synchronously-published frame-0.

This one reproduced directly: `test_sim_emits_commands_and_records_dataset` now checks frame-0 for both episodes (seeds 100 and 101), and unfixed, episode 1's first `mjSTATE_INTEGRATION` sample came back stepped (sim time 2.002 vs a fresh reset's 2.000) — the run loop's tiny physics `timestep` makes the collision near-certain on every multi-trial reset. With the guard both episodes record their post-reset scene exactly.

PR #442, opened 2026-06-19
Add the Robotiq 2F-85 gripper to the franka viewer model

vertix:
## Summary
- Graft the Robotiq 2F-85 onto the franka flange for the 3D viewer via `attach_robotiq_2f85`, shared by the live driver (`Robot._build_robot_meta`) and the bundled static model (`bundled_franka_model`). A single `grip` signal animates the open-chain 4-bar: driver/spring_link swing the finger in, coupler/follower counter-rotate to keep the pads parallel; left/right are mirrored so one positive `grip` closes both.
- Render robot visuals translucent white in the viewer (a material injected at URDF load) so the pose-direction markers stay visible through the arm and gripper.
- `REAL_ROBOT_TRANSFORM` now **backfills** the bundled model — recordings that carry their own driver-emitted model win; `SIM_ROBOT_TRANSFORM` **overwrites**, since older sim recordings store raw MuJoCo XML under `urdf`.
- Fix the commanded-trajectory color key (`command`) after the `robot_commands`->`robot_command` rename, restoring orange-target / cyan-state trajectories.
- Vendor the 2F-85 visual meshes (converted from MuJoCo Menagerie `robotiq_2f85`, scaled to metres, BSD-3). The two largest base meshes are decimated to fit the repo's 1000 KB asset limit.

## Test plan
- `uv run --locked pytest --no-cov` — full suite passes (584 passed, 7 skipped).
- Tested locally in the analysis viewer on real droid teleop and on `sim_pnp`: the FR3 + 2F-85 (and the sim panda) render translucent, the gripper animates from the recorded `grip`, and both the orange commanded and cyan state trajectories appear.


positronic/drivers/roboarm/__init__.py
```diff
@@ -16,23 +16,137 @@
 # Import command submodule to make it accessible as roboarm.command
 from . import command
 
+_FLANGE_LINK = 'link8'
```
vertix: Is is not clear, why all this code is in __init__.py, including `bundled_franka_model`.
It should move somewhere else

vertix: Fixed in 41c923a: moved the URDF/mesh model builders and the 2F-85 graft into a new `roboarm/models.py`. `__init__.py` now holds only the package contract — `State`, `RobotStatus`, and the `command` re-export.

The four consumers (`franka.py`, `cfg/ds/internal.py`, `simulator/mujoco/sim.py`, and the inference test) import from `roboarm.models` directly; no re-export shim left in the package root.

positronic/drivers/roboarm/models.py
```diff
@@ -0,0 +1,163 @@
...
+        link_el = ET.Element('link', name=link)
+        visual = ET.SubElement(link_el, 'visual')
+        if visual_xyz is not None:
+            ET.SubElement(visual, 'origin', xyz=visual_xyz, rpy='0 0 0')
+        ET.SubElement(ET.SubElement(visual, 'geometry'), 'mesh', filename=mesh)
+        joint_el = ET.Element('joint', name=joint or f'{link}_fixed', type='revolute' if axis else 'fixed')
+        ET.SubElement(joint_el, 'origin', xyz=xyz, rpy=rpy)
+        ET.SubElement(joint_el, 'parent', link=parent)
+        ET.SubElement(joint_el, 'child', link=link)
+        if axis is not None:
```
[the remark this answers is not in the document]
vertix: Confirmed and fixed in 4c6143f. Reproduced exactly — `DLSIKSolver(bundled_franka_model()['urdf'], ...)._model` failed with `mass and inertia of moving bodies must be larger than mjMINVAL` at `right_driver`, because `ik._prepare_spec` strips visuals/collisions and the gripper's revolute-joint links had only visuals.

Fix: each 2F-85 link now carries a nominal `<inertial>` (mass 0.01, diagonal inertia 1e-5) so the URDF is a valid MuJoCo model. The values are arbitrary — the viewer ignores inertials and the IK here is kinematic (FK/Jacobian), so they don't affect the arm solve. Verified the franka URDF now compiles, the 7 arm joints remain addressable, and `test_ik.py` + the inference integration tests pass.

positronic/drivers/roboarm/__init__.py
```diff
@@ -111,4 +66,4 @@ def ee_wrench(self) -> np.ndarray | None:
         return None
 
 
-__all__ = ['RobotStatus', 'State', 'bundled_franka_model', 'bundled_panda_model', 'command']
+__all__ = ['RobotStatus', 'State', 'command']
```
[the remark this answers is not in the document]
vertix: Declining this one. This repo's convention (CLAUDE.md) is to break internal imports cleanly rather than leave compat re-exports: *"Internal code breaks cleanly — no speculative compat shims. Before a migrate-everywhere change, grep for consumers that need the old form; alias or migrate only the real ones."*

I migrated every consumer in this PR — `franka.py`, `cfg/ds/internal.py`, `simulator/mujoco/sim.py`, and `tests/test_inference_integration.py` — and verified there are zero remaining root-level imports of these names anywhere in the repo (all file types) and none in the sibling `../gr00t` / `../openpi` checkouts. The full suite passes. There is no external published API here, so "downstream breakage on upgrade" doesn't apply.

A re-export alias would also reintroduce the package-root surface that this move was explicitly asked to remove (the PR's first review comment), so the clean break is the intended end state. Leaving this open for @vertix to weigh in.

positronic/drivers/roboarm/franka.py
```diff
@@ -147,11 +148,13 @@ def _build_robot_meta(robot) -> dict:
             stl = link.get('name', '') + '.stl'
             if stl in meshes and link.find('visual') is None:
                 ET.SubElement(ET.SubElement(link, 'visual'), 'geometry').append(ET.Element('mesh', filename=stl))
+        gripper = attach_robotiq_2f85(root, meshes)
         return {
             'urdf': ET.tostring(root, encoding='unicode'),
             'joint_names': _revolute_joint_names(urdf_xml),
             'meshes': meshes,
             'control_frame': 'end_effector',
+            'gripper': gripper,
```
vertix, line 159: Pls add a TODO to move it out the franka to the gripper driver and combine these two things together on the embodiment level.

vertix: Added in 507a81c:

```python
# TODO: The arm driver should not own the gripper. Move the 2F-85 model to the gripper driver
# (drivers/gripper) and compose the arm and gripper together at the embodiment level.
gripper = attach_robotiq_2f85(root, meshes)
```

positronic/cfg/ds/internal.py
```diff
@@ -48,7 +48,11 @@
 # key clash (Group prefers the earlier transform), so this alias is a no-op for them.
 _RENAME_ROBOT_COMMAND = Derive(**{'robot_command.pose': Get('robot_commands.pose', None)})
 
-REAL_ROBOT_TRANSFORM = Group(Derive(**_REAL_ROBOT_DERIVES), Identity(remove=_REMOVE_SIGNALS), _RENAME_ROBOT_COMMAND)
+# Real recordings carry the arm's own driver-emitted model, so the bundled one only backfills when
+# absent (Identity first wins on a clash). The sim delivers its model through the transform, not the
+# recording (older sim recordings store raw MuJoCo XML under `urdf`), so the bundled panda overwrites
+# (Derive first wins).
+REAL_ROBOT_TRANSFORM = Group(Identity(remove=_REMOVE_SIGNALS), Derive(**_REAL_ROBOT_DERIVES), _RENAME_ROBOT_COMMAND)
```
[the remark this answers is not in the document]
vertix: Confirmed and fixed in 008c4e1. You're right — `ROBOT_STATIC_META` (eval.py) is written into recorded episodes via `data_collection.py`, so pre-#440 recordings carry `pose_signals=['robot_state.ee_pose', 'robot_commands.pose']` in static. With `Identity` first, that stale value won and `_REMOVE_SIGNALS` stripped the plural signal, so the commanded trajectory wasn't logged.

Fix: split the derives so the signal-name pointers (`pose_signals`, `joint_signal`) always come from the transform (first in each Group, so they override stale statics), while the model payload still backfills for real (Identity before `_REAL_MODEL`) and overwrites for sim. Verified on synthetic episodes: a stale recording now yields canonical `pose_signals` while its own `urdf` still wins; empty real backfills the bundled franka; sim overwrites a bogus MJCF `urdf`. Full suite passes.

PR #451, opened 2026-06-23
Authenticate inference endpoints in-process; add `secure_remote`/`modal_remote` configs

vertix:
## Summary

Authenticated access to served inference endpoints, on by default. Auth is enforced **in the server**,
not at the Nebius ingress: a live test showed Nebius `--auth token` gates plain HTTP but **drops the
inference WebSocket upgrade** (`/api/v1/session` → 404 even with a valid token), so it can't protect
the actual inference path. Instead `VendorServer` validates the bearer token itself, and endpoints are
served open at the Nebius layer.

- **In-process auth (default on):** `VendorServer` reads `AUTH_TOKEN` from the environment and validates
  `Authorization: Bearer <token>` on the HTTP route (`/api/v1/models`) and the WebSocket routes
  (`/api/v1/session[/{model_id}]`) — constant-time compare, rejecting a missing/invalid token on the WS
  before `accept()`. Absent token ⇒ open server. Read once in the base class, so no per-vendor churn.
- **`serve.sh`:** injects the token as the container's `AUTH_TOKEN` via `--env-secret`
  (`NEBIUS_AUTH_TOKEN_SECRET`, default `positronic-serverless-inference-token`), replacing `--auth token`.
  `NEBIUS_AUTH_TOKEN_SECRET=` serves an open endpoint; `e2e.sh` opts out for its smoke check.
- **Two client policy configs** over `remote` that attach the auth header from the environment, so
  callers pass only `--policy.host`:
  - `.secure_remote` → `Authorization: Bearer` from `POSITRONIC_INFERENCE_TOKEN` (our Nebius endpoints, HTTP `:8000`)
  - `.modal_remote` → `Modal-Key`/`Modal-Secret` from `MODAL_PROXY_TOKEN_ID`/`MODAL_PROXY_TOKEN_SECRET`, TLS `:443`
  - Each raises immediately if its environment variable is unset.
- **Docs:** the bearer-token secret joins the one-time setup; `workflows/nebius/README.md` and the
  `remote-training` skill describe the in-process-auth flow.
- Also includes a small unrelated `CLAUDE.md` tweak (allow autonomous commits; prefer fewer test files).

## Test plan

Verified:
- **Auth logic (local):** FastAPI `TestClient` matrix — authed server rejects missing/wrong token (HTTP 401,
  WS rejected) and accepts a valid token (HTTP 200, WS → ready); open server passes through. Regression
  cases added to `test_vendor_server.py` exercise the same over real uvicorn + websockets.
- **Full suite:** `pytest --no-cov` → 598 passed, 7 skipped. `ruff check`/`format` clean; `bash -n` clean.
- **Live Nebius control:** deployed an open endpoint — raw WS upgrade → 101 and real
  `InferenceClient.new_session()` → ready. A prior authed (`--auth token`) endpoint showed tokenless
  HTTP → 401, authed HTTP → 200, but authed WS → 404 — confirming the ingress drops the WS upgrade.

Not verified (needs an image rebuild + deploy): the `--env-secret AUTH_TOKEN=` injection inside a live
Nebius container — it mirrors the existing, working `--env-secret AWS_ACCESS_KEY_ID=` line.

### DIFF
--- a/positronic/cfg/policy.py
+++ b/positronic/cfg/policy.py
@@ -1,3 +1,5 @@
+import os
+
 import configuronic as cfn
 import pos3
 
@@ -91,6 +93,32 @@ def remote(
             return rec.tap('raw').wrap(policy)
         return (rec.tap('raw') | codec | rec.tap('server')).wrap(policy)
     return codec.wrap(policy) if codec else policy
+
+
+@cfn.config(env_var='POSITRONIC_INFERENCE_TOKEN')
+def bearer_headers(env_var: str):
+    token = os.environ.get(env_var)
+    if not token:
+        raise ValueError(f'{env_var} is not set; export the endpoint auth token before running inference.')
+    return {'Authorization': f'Bearer {token}'}
+
+
+@cfn.config(id_env='MODAL_PROXY_TOKEN_ID', secret_env='MODAL_PROXY_TOKEN_SECRET')
+def modal_headers(id_env: str, secret_env: str):
+    key, secret = os.environ.get(id_env), os.environ.get(secret_env)
+    missing = [name for name, value in [(id_env, key), (secret_env, secret)] if not value]
+    if missing:
+        raise ValueError(f'Modal proxy credentials not set; export {" and ".join(missing)} before running inference.')
+    return {'Modal-Key': key, 'Modal-Secret': secret}
+
+
+# Our own Nebius endpoints answer plain HTTP on :8000; for an endpoint served with Nebius
+# `--auth token`, the bearer token is the access control. Front it with TLS instead by adding
+# --policy.secure=True --policy.port=443.
+secure_remote = remote.override(headers=bearer_headers)
+
+# Modal fronts every endpoint with TLS on :443, gated on its proxy key/secret pair.
+modal_remote = remote.override(port=443, secure=True, headers=modal_headers)
 
 
 @cfn.config(
--- a/positronic/cfg/policy.py
+++ b/positronic/cfg/policy.py
@@ -112,8 +112,9 @@ def modal_headers(id_env: str, secret_env: str):
     return {'Modal-Key': key, 'Modal-Secret': secret}
 
 
-# Our own Nebius endpoints answer plain HTTP on :8000 and validate the bearer token in-process.
-# Front them with TLS instead by adding --policy.secure=True --policy.port=443.
+# Our own Nebius endpoints are bare-IP plain HTTP on :8000 (Nebius serves no TLS), so the bearer token
+# rides the client-to-endpoint path in cleartext and is capturable there. Harden by fronting the endpoint
+# with a TLS terminator and adding --policy.secure=True --policy.port=443.
 secure_remote = remote.override(headers=bearer_headers)
 
 # Modal fronts every endpoint with TLS on :443, gated on its proxy key/secret pair.
--- a/positronic/cfg/policy.py
+++ b/positronic/cfg/policy.py
@@ -112,9 +112,8 @@ def modal_headers(id_env: str, secret_env: str):
     return {'Modal-Key': key, 'Modal-Secret': secret}
 
 
-# Our own Nebius endpoints answer plain HTTP on :8000; for an endpoint served with Nebius
-# `--auth token`, the bearer token is the access control. Front it with TLS instead by adding
-# --policy.secure=True --policy.port=443.
+# Our own Nebius endpoints answer plain HTTP on :8000 and validate the bearer token in-process.
+# Front them with TLS instead by adding --policy.secure=True --policy.port=443.
 secure_remote = remote.override(headers=bearer_headers)
 
 # Modal fronts every endpoint with TLS on :443, gated on its proxy key/secret pair.
--- a/positronic/offboard/vendor_server.py
+++ b/positronic/offboard/vendor_server.py
@@ -1,7 +1,9 @@
 """Base class for vendor inference servers (GR00T, OpenPI, DreamZero, LeRobot)."""
 
 import asyncio
+import hmac
 import logging
+import os
 import time
 from abc import ABC, abstractmethod
 from collections.abc import Callable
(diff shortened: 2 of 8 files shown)


workflows/nebius/README.md
```diff
@@ -267,6 +267,92 @@ bash workflows/nebius/stop.sh my-act-demo
...
+export POSITRONIC_INFERENCE_TOKEN=$(nebius mysterybox payload get-by-key \
+  --secret-id "$SECRET_ID" --key AUTH_TOKEN --format json | jq -r '.data.string_value')
+
+uv run positronic-inference sim \
+  --policy=.secure_remote \
+  --policy.host=<endpoint-ip> \
+  --output_dir=.data/inference/<run-name>/
+```
+
+### A Modal endpoint
```
vertix: Modal is not related to nebius, remove completely

workflows/nebius/README.md
```diff
@@ -267,6 +267,92 @@ bash workflows/nebius/stop.sh my-act-demo
 To pause an endpoint without releasing its static IP (useful if you want to reuse the same IP
 later), use `nebius ai endpoint stop <id>` directly — `start` resumes it.
 
+## Authenticated inference
```
vertix, line 286: This is too long. Modal has nothing to do with nebius.

workflows/nebius/README.md
```diff
@@ -267,6 +267,92 @@ bash workflows/nebius/stop.sh my-act-demo
...
+pair. Both raise immediately if their environment variable is unset, rather than sending an
+unauthenticated request.
+
+A bare value retrieved from MysteryBox comes wrapped in a small JSON envelope — extract the raw
+token with `--format json | jq -r '.data.string_value'` (not `--format text`, which returns the
+whole envelope).
+
+### Our own Nebius endpoint
+
+**One-time:** create a MysteryBox secret holding the bearer token. The payload key must be
```
vertix: There's existing section in the top of the documents related to secrets, move installation there. Just the esssense, make it concise

workflows/nebius/README.md
vertix: Condensed the section and removed Modal entirely; this README is now Nebius-only.

workflows/nebius/README.md
vertix: Removed the Modal section and the `.modal_remote` table row from this README. The `.modal_remote` config stays in the code, just not documented in the Nebius workflow.

workflows/nebius/README.md
vertix: Condensed: dropped the policy-config table and the duplicate secret-fetch. Secret creation already lives in the top One-time setup; this section now just loads the token and runs. Left the per-run token export here rather than in setup, but glad to move it up if you prefer.

workflows/nebius/README.md
```diff
@@ -267,6 +283,73 @@ bash workflows/nebius/stop.sh my-act-demo
...
+export POSITRONIC_INFERENCE_TOKEN=$(nebius mysterybox payload get-by-key \
+  --secret-id "$SECRET_ID" --key AUTH_TOKEN --format json | jq -r '.data.string_value')
+
+uv run positronic-inference sim \
+  --policy=.secure_remote \
+  --policy.host=<endpoint-ip> \
+  --output_dir=.data/inference/<run-name>/
+```
+
+Rotate the token by adding a new primary version (clients re-read it on their next run):
```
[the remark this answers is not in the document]
vertix: Fixed in 3b7a946: the rotation note now states that a running endpoint keeps validating the token it read at container start, and you must re-run `serve.sh` for each live endpoint after rotating or new clients hit 401/1008 against the old token.

positronic/offboard/vendor_server.py
```diff
@@ -146,10 +148,17 @@ def __init__(
 
         self.metadata: dict[str, Any] = {}
 
+        # Bearer token gating every endpoint, injected as the container's AUTH_TOKEN env var by
+        # serve.sh. Absent -> the server is open. Validated in-process because Nebius `--auth token`
+        # gates HTTP but does not proxy the inference WebSocket.
+        self._auth_token = os.environ.get('AUTH_TOKEN') or None
```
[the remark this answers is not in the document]
vertix: Fixed in c7d9314: an absent `AUTH_TOKEN` still means open (the intended opt-out, `NEBIUS_AUTH_TOKEN_SECRET=`), but a set-but-empty `AUTH_TOKEN` now raises at startup so the endpoint never comes up serving open. The check is an explicit `raise` rather than an `assert` so `-O` can't strip it. `_authorized` already keys open mode on `is None`, so the contract is now: `None` -> open, non-empty -> enforced, empty -> can't construct. Added `test_empty_auth_token_fails_closed`.

positronic/cfg/policy.py
```diff
@@ -93,6 +95,31 @@ def remote(
...
+    key, secret = os.environ.get(id_env), os.environ.get(secret_env)
+    missing = [name for name, value in [(id_env, key), (secret_env, secret)] if not value]
+    if missing:
+        raise ValueError(f'Modal proxy credentials not set; export {" and ".join(missing)} before running inference.')
+    return {'Modal-Key': key, 'Modal-Secret': secret}
+
+
+# Our own Nebius endpoints answer plain HTTP on :8000 and validate the bearer token in-process.
+# Front them with TLS instead by adding --policy.secure=True --policy.port=443.
+secure_remote = remote.override(headers=bearer_headers)
```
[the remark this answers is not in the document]
vertix: Valid concern, but the literal remedy is not applicable to these endpoints, and I want to leave this open for a human call rather than resolve it.

Nebius serverless AI endpoints are bare public IPs over plain HTTP with no TLS (confirmed in the docs: the endpoint exposes a `public_ip`, every example is `http://<ip>`, and there is no certificate/hostname; a live test also showed the `--auth token` proxy drops the inference WebSocket upgrade). Consequences:

- Requiring `secure=True` on `.secure_remote` would make it speak `wss://`/`https://` to an endpoint that only serves plain HTTP, so it could no longer reach the endpoints it exists for.
- "Don't use bearer-on-http" isn't an option either: bearer-on-http is the only mechanism that authenticates a Nebius endpoint at all, which is the entire purpose of this change.

So the real hardening is fronting the endpoint with a TLS terminator (hostname + cert), which is separate infrastructure work, out of scope for this PR. What I did land (139af3c) is to make the exposure explicit at the config site rather than silently recommend cleartext bearer. The token is also rotatable (see the README rotation recipe). Leaving this open as a documented known limitation pending a decision on whether to add TLS fronting.

vertix:
Superseded: in-process bearer auth shipped in `positronic/offboard/server.py` (`AUTH_TOKEN_ENV`, `AUTH_HEADER`, WS auth before `accept()`) with the secret plumbing in `positronic/utils/nebius.py` and the `.authed_remote` client config. The `.modal_remote` config is the only part not carried over.

PR #450, opened 2026-06-22
Add a remote env-server boundary (`RemoteEnvControlSystem` + `EnvAdapter`)

vertix:
## Summary

Adds the remote env-server boundary on top of the merged harness / control-loop refactor (#448).

- New `positronic/simulator/env_server/` module: a socket `proxy`/`client`/`protocol`, a `server`, and an `EnvAdapter` translating env steps to and from the wire. `RemoteEnvControlSystem` stays a dumb translator; the adapter owns trajectory playing and the canonical↔raw mappings.
- Reuses the harness's frame-0 / reset-as-flag / record-ordering contract across the socket boundary, so a remote env records identically to a local one.
- Tests cover the boundary itself: wire round-trip parity, the proxy's run loop (frame-0 + free-run), the real `stack_cubes` wrapper end-to-end over the socket, and error-frame propagation.

## Test plan

`uv run --locked pytest --no-cov` → 595 passed, 7 skipped (full suite, including the env-server boundary tests).

Supersedes #438, which auto-closed when its base branch (`harness-control-loop`) was deleted on the #448 merge.

PR #452, opened 2026-06-23
Add LIBERO as a remote env-server embodiment

vertix:
## Summary

- Adds **LIBERO** as a remote embodiment behind the env-server boundary (#450): `LiberoEnv` runs in its own Python 3.10 interpreter (bootstrapped by `uv run --no-project`, since LIBERO pins ≤3.10), driven over the socket by positronic's Harness — latency-honest, no shared venv.
- `LiberoEnv` maps every positronic command (Cartesian pose / joint position / joint velocity / hold) into the active robosuite controller's normalized action — the exact inverse of robosuite's `scale_action` + `set_goal` — bridging pose↔joint with FK and damped-least-squares IK on the MuJoCo site Jacobian. It reports the eef pose in the **grip-site frame it controls** (robosuite's raw `eef_quat` is the hand-body frame, a fixed offset) so observation and command frames stay consistent for absolute-pose control.
- The env-server proxy now connects **lazily on first reset** (so positronic launches the server subprocess first), with a longer client deadline for the 3.10 env build.
- **Validation:**
  - `validate.py` — checks the transform is byte-identical to robosuite's own control path (drive `set_goal` forward, invert via `_arm_action`, assert recovery).
  - `e2e.py` — replays recorded LIBERO demos through the **real socket** and asserts task success. Non-circular: each waypoint is anchored on the wire observation's eef pose, so a wrong quaternion order / frame / scale makes the demo diverge and fail. Episodes come from a tiny committed `.npz` fixture (`make_fixture.py`), not the multi-GB benchmark.
- **CI:** a path-filtered lane (`libero-e2e.yaml`) runs the e2e with software rendering (`MUJOCO_GL=osmesa`) on a stock GPU-free runner.

Out of scope (next PR): the closed-loop openpi `pi05_libero` policy, and the remaining observation conventions in `adapter.py` (camera image flip, gripper-qpos → [0,1]).

## Test plan

- `uv run --locked pytest --no-cov` — full suite green: **595 passed, 8 skipped**. The env-server boundary change is covered by the existing `env_server` tests; the LIBERO e2e self-skips until its fixture is committed.
- The command transform was verified against robosuite 1.4.1 source and numerically (the `e2e.py` compose inverts the server's `_arm_action` to float precision). It has **not** been executed on a LIBERO box — `validate.py` and `e2e.py` need LIBERO's 3.10 env + a demo fixture, which aren't in this CI.
- To make the e2e a live gate: run `make_fixture.py` once on a LIBERO box, commit the `.npz`; the `libero-e2e` lane then replays a demo through the full env server on every simulator change.


positronic/simulator/libero/launcher.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    port = _free_port()
+    proc = _spawn(suite, task_id, camera_resolution, control_mode, host, port)
+    try:
+        yield host, port
+    finally:
+        _terminate(proc)
+
+
+class LiberoServer(pimm.ControlSystem):
```
vertix: Wait, this is smth new. Why do we run it as control system? Second, why can't we use existing server that we have in ../env_server?

positronic/cfg/eval/sim/libero.py
```diff
@@ -0,0 +1,60 @@
...
+    privileged = {'sim_state': Observation(proxy.privileged['sim_state'], None)}
+    task = Task(
+        instruction=instruction, timeout=timeout, privileged=privileged, seed=seed, reset=proxy.reset, done=proxy.done
+    )
+    return Eval(embodiment, task)
+
+
+# TODO: ``instruction`` must match the LIBERO task's own ``language`` (the server reports it in frame-0
+# meta) — verify against the suite when wiring a real policy.
+spatial = _libero_eval.override(
```
vertix: Pls explain why we have this configuration, I don't understand. And why can't we just use _libero_eval?
If we have other suites, we should have configs for all of them.
What is task_id. The doc should have a simple explanation and link to the libero website/repo

positronic/simulator/libero/tests/test_e2e.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@pytest.mark.skipif(
+    not os.path.exists(_FIXTURE),
+    reason='no e2e fixture committed yet — generate it on a LIBERO box with make_fixture.py',
+)
+@pytest.mark.timeout(900)  # the env server bootstraps its 3.10 deps on first run, which can take minutes
+def test_demo_replay_reaches_success():
+    rate = run_replay(_FIXTURE, suite='libero_spatial', task_id=0)
+    assert rate >= 0.66, f'demo replay success rate {rate:.2f} — env server likely broken'
```
vertix: Why 0.66? Should not it be 1.0 on the prerecorded track?

positronic/simulator/libero/validate.py
```diff
@@ -0,0 +1,129 @@
...
+# requires-python = "==3.10.*"
+# dependencies = [
+#     "libero @ git+https://github.com/Lifelong-Robot-Learning/LIBERO.git",
+#     "robosuite==1.4.1",
+#     "numpy<2",
+#     "msgpack",
+#     "websockets",
+# ]
+# ///
+"""On-box validation for ``LiberoEnv``'s command transform — runs inside LIBERO's own 3.10 interpreter.
```
vertix, line 24: As far as I understand, this is our first test, not e2e test that tests the full integration.
If so, should we reomve it? Does it carry any value? If it carries extra value, should we include it in CI?

positronic/simulator/libero/env.py
```diff
@@ -0,0 +1,274 @@
...
+        task_suite = benchmark.get_benchmark_dict()[self._suite_name]()
+        task = task_suite.get_task(self._task_id)
+        self._init_states = task_suite.get_task_init_states(self._task_id)
+        bddl = pathlib.Path(get_libero_path('bddl_files')) / task.problem_folder / task.bddl_file
+        # OSC_POSE is the controller LIBERO configures for its own policies (its ``env_wrapper`` hands the
+        # controller *name* to ``load_controller_config``); operational-space control maps a per-step
+        # end-effector pose delta to joint motion — LIBERO ships no IK of its own. The joint modes select
+        # robosuite's joint controllers. Pass the name, not a config dict, so the wrapper loads it as LIBERO does.
+        controller_names = {'ee': 'OSC_POSE', 'joint': 'JOINT_POSITION', 'joint_delta': 'JOINT_VELOCITY'}
+        # TODO: verify against a live LIBERO install — ``env.control_freq`` and ``sim.get_state().flatten()``
```
vertix: Can you verify and close it?

positronic/simulator/libero/env.py
```diff
@@ -0,0 +1,274 @@
...
+        # are env internals unverifiable in 3.11.
+        self._env = OffScreenRenderEnv(
+            bddl_file_name=str(bddl),
+            controller=controller_names[self._control_mode],
+            camera_heights=self._resolution,
+            camera_widths=self._resolution,
+        )
+        self._meta = {'suite': self._suite_name, 'task_id': self._task_id, 'task': task.language}
+        self._control_dt = 1.0 / self._env.env.control_freq
+        robot = self._env.env.robots[0]
```
vertix, line 135: Just for my understanding, are there multiple robots, or just one?

positronic/simulator/libero/env.py
```diff
@@ -0,0 +1,272 @@
...
+            raw = self._env.set_init_state(self._init_states[seed % len(self._init_states)])
+        return {'obs': self._observe(raw), 'meta': self._meta, 'control_dt': self._control_dt}
+
+    def step(self, action: dict[str, Any]) -> dict[str, Any]:
+        arm = self._arm_action(action['command'])
+        # positronic grip in [0, 1] maps to robosuite's [-1, 1]; robosuite's PandaGripper opens at -1 and closes
+        # at +1, so grip=1 closes.
+        grip = action['grip'] * 2.0 - 1.0
+        raw, _reward, done, _info = self._env.step(np.concatenate([arm, [grip]]).tolist())
+        return {'obs': self._observe(raw), 'done': bool(done), 'control_dt': self._control_dt}
```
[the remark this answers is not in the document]
vertix: Agree, we should clearly say that the benchmark "released" success in our payload that we send via `done` channel

positronic/simulator/libero/env.py
```diff
@@ -0,0 +1,274 @@
...
+        data.qpos[self._qpos_idx] = q
+        mujoco.mj_forward(self._model, data)
+        pos = np.array(data.site_xpos[self._eef_site_id])
+        rot = np.array(data.site_xmat[self._eef_site_id]).reshape(3, 3)
+        return pos, rot
+
+    def _ik(self, target_pos: np.ndarray, target_rot: np.ndarray) -> np.ndarray:
+        # Damped-least-squares differential IK on the same MuJoCo site Jacobian OSC uses (no pybullet), iterated
+        # on a scratch ``MjData`` from the current joints.
+        # TODO (verify on a LIBERO box): the iteration count / damping and the ``get_pose_error`` orientation frame.
```
vertix: Please do

positronic/simulator/libero/env.py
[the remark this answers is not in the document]
vertix: Declining the code change, but agreeing on the substance. `done` is already LIBERO success: `BDDLBaseDomain.step` (`libero/libero/envs/bddl_base_domain.py:800-808`) calls `super().step()` and then overrides `done = self._check_success()`. So `bool(done)` is `True` exactly on task success, and at the step limit `done = _check_success()` is `False` (no false success). Documented this at the source in 5624972 so the `done` payload's meaning is explicit. (`self._env` is the `OffScreenRenderEnv` wrapper, so the method would be `self._env.env._check_success()` anyway — but moot, since `done` carries it.)

positronic/cfg/eval/sim/libero.py
```diff
@@ -0,0 +1,60 @@
...
+    }
+    commands = {
+        'robot_command': Command(proxy.commands['robot_command'], roboarm_command.Reset(), Serializers.robot_command),
+        'target_grip': Command(proxy.commands['target_grip'], 0.0, None),
+    }
+    embodiment = Embodiment(
+        descriptor='remote.libero.franka',
+        observations=observations,
+        commands=commands,
+        static_meta=dict(ROBOT_STATIC_META),
```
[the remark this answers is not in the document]
vertix: Fixed in 5624972. LIBERO drives the same Franka Panda as the native sim, so the eval's `static_meta` now carries `bundled_panda_model()` (URDF + meshes + joint_names + control_frame) — the same model the native path emits via `robot_meta`. Added on the positronic side (the cfg) since the 3.10 server can't import positronic to emit it. One thing to confirm once LIBERO episodes are actually recorded (the closed-loop eval PR): the bundled `control_frame` (`end_effector`) vs robosuite's grip site may carry a small fixed offset for offline IK; the arm geometry / joints are unaffected.

positronic/simulator/libero/launcher.py
vertix: We do reuse it: `env.py` runs `EnvServer(env, ...).serve_forever()` — the `../env_server` WebSocket server — inside LIBERO's 3.10 interpreter. `LiberoServer` here is the *positronic-side launcher*: it spawns that 3.10 subprocess (`uv run --no-project env.py`) and owns its lifetime. It is a `ControlSystem` only so the pimm World tears the subprocess down when the run ends — it holds no ports and does no per-tick work (`RemoteEnvControlSystem` does the socket talking). There's also a plain `serve_libero` context-manager form in `launcher.py` for non-World callers like the e2e.

positronic/cfg/eval/sim/libero.py
vertix: Improved the docstring in 5624972: a LIBERO *suite* is a set of related tasks; `task_id` selects one task within it (a fixed scene + language goal), with a link to the LIBERO repo. `_libero_eval` can't be used directly because `suite` and `instruction` are unbound — `spatial` is the concrete `.override` that binds them, and `instruction` must match that task's own `language` (which the server reports in frame-0 meta; see the TODO at the bottom of the file).

On more suites: each `(suite, task)` is a named override like `spatial`. I kept just `spatial` as the first one rather than enumerating every suite×task (each needs its own task-specific instruction). Happy to add the specific ones you want to run — or, cleaner, derive `instruction` straight from the server meta so named configs don't need to hardcode it. Which would you prefer?

positronic/simulator/libero/tests/test_e2e.py
vertix: Agreed — fixed in 5624972, the threshold is now `== 1.0`. The fixture holds prerecorded *successful* tracks, so a faithful replay must reproduce every one; anything below 1.0 means the env diverged.

positronic/simulator/libero/validate.py
vertix: It isn't redundant with the e2e. The demo-replay e2e exercises only the `ee`/`OSC_POSE` path (the demos are OSC). `validate.py` additionally covers the `joint`/`JOINT_POSITION` and `joint_delta`/`JOINT_VELOCITY` control modes, plus FK identity and the IK round-trip — none of which the replay touches. It's also the cheap check (no fixture, no rendering, no demos), so it's the fast guard for the transform algebra. It needs LIBERO's 3.10 env, so it belongs in the same on-box / `libero-e2e` lane as the e2e, not the default CI. I'd keep it — but if you'd rather drop the joint-mode coverage for now, I can remove it.

positronic/simulator/libero/env.py
vertix: Verified and removed in 5624972: `control_freq` is a standard robosuite base-env attribute (`robosuite/environments/base.py:119`) and `sim.get_state().flatten()` is the standard MjSim state API — both stable robosuite, nothing that needed an on-box check.

positronic/simulator/libero/env.py
vertix: Just one. LIBERO tasks are single-arm (the Franka Panda), so `robots[0]` is the only robot. `robots` is robosuite's general multi-robot list, but LIBERO never populates more than one.

positronic/simulator/libero/env.py
vertix: Done in 5624972. Verified the concretely-checkable part — `get_pose_error` (`robosuite/utils/transform_utils.py`) returns the world-frame rotation error, which matches the world-frame site Jacobian the DLS IK uses — and `validate.py`'s `_fk(_ik(pose))` round-trip covers convergence (the iteration count / damping). Replaced the TODO with that note. (The IK only feeds the joint-control-mode Cartesian cells; the primary `ee`/OSC path doesn't use it.)

positronic/simulator/libero/env.py
```diff
@@ -0,0 +1,272 @@
+# /// script
+# requires-python = "==3.10.*"
+# dependencies = [
+#     "libero @ git+https://github.com/Lifelong-Robot-Learning/LIBERO.git",
```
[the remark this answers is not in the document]
vertix: Valid, verified: LIBERO's `setup.py` has `install_requires=[]`, so the deps in its `requirements.txt` (`bddl`, `hydra-core`, `robomimic`, `transformers`, `gym`, `matplotlib`, ...) are **not** pulled by `libero @ git+...`. Importing `libero.libero.envs` will `ModuleNotFoundError` on a clean runner.

The fix is to add them to the PEP 723 list — but LIBERO's pins (`robosuite==1.4.0`, `numpy==1.22.4`, `transformers==4.21.1`, `gym==0.25.2`, ...) have to reconcile with our `robosuite==1.4.1` + `mujoco`, and the exact resolvable/importable set is only confirmable by an actual install on a LIBERO box. Since CI can't validate it (the lane skips without a fixture), I'm not guessing a dep set blind — leaving this for the on-box bring-up, where the e2e run pins the working set. Tracked here.

positronic/simulator/libero/env.py
```diff
@@ -0,0 +1,272 @@
...
+import argparse
+import pathlib
+from typing import Any
+
+import mujoco
+import numpy as np
+from robosuite.utils.transform_utils import get_pose_error, make_pose, mat2quat, quat2axisangle
+from server import EnvProtocol, EnvServer
+
+from libero.libero import benchmark, get_libero_path
```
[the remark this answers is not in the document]
vertix: Fixed in 65a5019 (not yet validated on a box). LIBERO's package `__init__` calls `input()` for the dataset path when `~/.libero/config.yaml` is absent, which would block this unattended subprocess. The server now shims `builtins.input` to answer the default before importing `libero.libero`, so LIBERO writes its default (package-relative) config non-interactively and never reads stdin. Leaving the thread open until a clean on-box e2e run confirms the full startup.

positronic/simulator/libero/env.py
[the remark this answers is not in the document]
vertix: Fixed in 6b917a5, and validated end-to-end on macOS. Two problems, both real:

- **Undeclared deps:** env.py's PEP 723 block now declares LIBERO's actual runtime deps — `torch<2.6` (2.6 flips `torch.load` to `weights_only=True`, which breaks LIBERO's pickled init-states), `bddl==1.0.1`, `gym==0.25.2`, `hydra-core`, `easydict`, `einops`, `cloudpickle`, `future`, `matplotlib`, `pyyaml` — alongside `robosuite`/`mujoco`/`numpy`.
- **`libero` isn't wheel-installable at all** (this is deeper than the deps): it ships `libero/` as a PEP 420 namespace package with no top-level `__init__.py`, so `find_packages()` builds an empty wheel and `import libero` fails even with every dep present. The launcher now clones the pinned LIBERO source into `~/.cache/positronic/libero` and puts the repo root on the subprocess `PYTHONPATH` — the bddl/init-state/asset files the env reads live in the tree, not a wheel.

The committed demo-replay e2e now actually builds the 3.10 env, clones LIBERO, and replays to task success (`1 passed`).

positronic/simulator/libero/env.py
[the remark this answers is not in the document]
vertix: Fixed, in two layers: 65a5019 answers the first-run stdin prompt non-interactively (the `builtins.input` shim before the libero imports), and 6b917a5 sets a prepared `LIBERO_CONFIG_PATH` beside the cached checkout — exactly your suggestion — so the config is hermetic and can't go stale against a `~/.libero` from an earlier, differently-located clone. Exercised on a clean config dir as part of the e2e bring-up.

positronic/cfg/eval/sim/libero.py
```diff
@@ -0,0 +1,70 @@
...
+    embodiment = Embodiment(
+        descriptor='remote.libero.franka',
+        observations=observations,
+        commands=commands,
+        # LIBERO drives the same Franka Panda as the native sim, so recordings carry the same model (URDF +
+        # meshes + joint names + control frame) for the 3D viewer and offline IK, supplied here since the 3.10
+        # server can't import positronic to emit it via ``robot_meta``.
+        static_meta={**ROBOT_STATIC_META, **bundled_panda_model()},
+        meta_source=proxy.robot_meta,
+        control_systems=(server, proxy),
```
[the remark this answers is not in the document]
vertix: Fixed in 714bbc7 — went with the tolerant-close option rather than reordering, because reordering wouldn't reliably help. The harness doesn't finalize `control_systems` in tuple order: `World.interleave` (pimm/world.py) never closes the generators explicitly; each exits via `StopIteration` when `should_stop` flips, and which one is driven first is decided by per-loop **sleep timing** (`LiberoServer` sleeps 1.0s, the proxy sleeps the env's `control_dt`), not the tuple index. So no ordering guarantees proxy-before-server.

`EnvConnection.close()` now treats the close handshake as best-effort: a `ConnectionClosed` while sending `close` or awaiting the ack means the peer is already gone, and the socket is closed regardless — so a rollout can't fail during shutdown whatever order the two control systems finalize in.

pyproject.toml
```diff
@@ -138,6 +138,7 @@ testpaths = [
     "positronic/offboard/tests",
     "positronic/policy/tests",
     "positronic/simulator/env_server/tests",
+    "positronic/simulator/libero/tests",
```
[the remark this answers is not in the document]
vertix: Fixed in 8aa98c5 — dropped `positronic/simulator/libero/tests` from the default `testpaths`. The dedicated `libero-e2e.yaml` lane already runs it by explicit path (`pytest positronic/simulator/libero/tests`), so a plain `uv run pytest` no longer collects the heavy replay. Verified: default collection now finds 0 libero tests, the explicit path still finds `test_demo_replay_reaches_success`.

positronic/simulator/libero/adapter.py
```diff
@@ -0,0 +1,89 @@
...
+        return {'command': _wire_command(cmd), 'grip': float(self._held.get('target_grip', 0.0))}
+
+    def observations(self, raw_obs: dict[str, Any]) -> dict[str, Any]:
+        # The env reports the eef pose in the grip-site frame it controls; ``eef_quat`` is scalar-last (xyzw,
+        # from ``mat2quat``), so ``from_quat_xyzw`` is the matching decode.
+        ee_pose = geom.Transform3D(raw_obs['eef_pos'], geom.Rotation.from_quat_xyzw(raw_obs['eef_quat']))
+        state = MujocoFrankaState()
+        state.encode(raw_obs['joint_pos'], raw_obs['joint_vel'], ee_pose)
+        # TODO (verify on a LIBERO box): map the two Panda finger qpos to a single [0, 1] openness.
+        obs: dict[str, Any] = {'robot_state': state, 'grip': float(np.sum(raw_obs['gripper_qpos']))}
```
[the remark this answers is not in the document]
vertix: Valid, and it's the gap the TODO on that line marks (`np.sum` of the two finger qpos cancels for the open Panda). Deferring it on purpose: nothing in this PR consumes the observed `grip` — there's no policy wired yet, and the e2e derives gripper commands from the demo's recorded actions, not from this observation. The correct finger-qpos→[0,1] mapping (and the camera flip on the next line) will be implemented and verified on a LIBERO box in the closed-loop-policy PR, which is where `grip` first gets a live consumer. Leaving this open as the tracked follow-up.

positronic/simulator/libero/env.py
```diff
@@ -0,0 +1,291 @@
...
+        # All-to-all: each command becomes the physical pre-scale quantity the active controller's set_goal adds to
+        # the current setpoint, then ``_normalize`` inverts the controller's scaling. The pose<->joint cells use
+        # FK/IK on the site Jacobian; the within-joint cells use the control period.
+        match (self._control_mode, command['type']):
+            case ('ee', 'cartesian'):  # OSC_POSE: world-frame pose error
+                physical = self._pose_error(*_unpack_pose(command['pose']))
+            case ('ee', 'joint_pos'):
+                physical = self._pose_error(*self._fk(command['q']))
+            case ('ee', 'joint_vel'):
+                physical = self._pose_error(*self._fk(self._cur_q() + command['dq'] * self._control_dt))
```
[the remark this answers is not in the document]
vertix: Fixed in cd2bf01, and verified on a LIBERO box. You're right: `sim.py` applies `JointDelta` as `q + dq` (a per-step delta), not a rate. The bridge now applies `dq` directly for the OSC_POSE and JOINT_POSITION cells, and divides by the control period only for the JOINT_VELOCITY controller (`joint_delta` mode), which is the one place rad/s is wanted.

Confirmed against robosuite's own control path: `validate.py` (now runnable — its deps were the same empty-wheel `libero @ git+`) drives each controller's `set_goal` forward and inverts via `_arm_action`, and all three inverses pass byte-exact (`joint_vel inverse: OK, atol 1e-9`), plus FK/IK. (That run also caught — and I fixed — a harness bug where the check read a stale controller `ee_pos` after reset; a real step refreshes it each turn, which is why the e2e was unaffected.)

positronic/simulator/libero/launcher.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def _ensure_libero_src() -> Path:
+    """Clone LIBERO once into the cache and return its repo root for ``PYTHONPATH``.
+
+    LIBERO declares ``install_requires=[]`` and ships ``libero`` as a PEP 420 namespace package (no top-level
+    ``__init__.py``), so ``find_packages`` builds an empty wheel — it is importable only with the repo root on
+    ``sys.path``, and the bddl/init-state/asset files the env reads live in the repo tree, not a wheel.
+    """
+    if not (_LIBERO_SRC / 'libero' / 'libero' / '__init__.py').exists():
```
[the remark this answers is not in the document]
vertix: Fixed in f17fb55. `_ensure_libero_src` now clones only when `.git` is absent, then on every call reads `git rev-parse HEAD` and, if it differs from `_LIBERO_COMMIT`, fetches and checks out the pin. So a cache cloned earlier (or by hand) at another revision is brought to the pinned commit before the path is returned, rather than imported as-is. Verified the e2e still passes against the existing (already-pinned) cache.

positronic/simulator/libero/adapter.py
```diff
@@ -0,0 +1,89 @@
...
+                    self._held.pop(name, None)
+            for value in player.advance(now_ns):
+                self._held[name] = value
+        # The server maps the held command into the active controller's action. Reset/Recover have no robosuite
+        # action, so they forward as a hold.
+        cmd = self._held.get('robot_command')
+        if isinstance(cmd, roboarm_command.Reset | roboarm_command.Recover):
+            self._held.pop('robot_command')
+            cmd = None
+        return {'command': _wire_command(cmd), 'grip': float(self._held.get('target_grip', 0.0))}
```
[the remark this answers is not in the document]
vertix: Fixed in bce925b. The adapter now keeps the last commanded grip in `self._grip` and re-sends it when the trajectory is cancelled, instead of falling back to `0.0` (which env.py maps to `-1`, opening the gripper). It updates `_grip` only while a grip waypoint is held, so a cancelled / robot-only chunk freezes the gripper where it is — matching how the arm forwards a `hold` on cancel. Initial state (before any grip command) is still `0.0`.

positronic/simulator/libero/adapter.py
```diff
@@ -0,0 +1,95 @@
...
+            adapter = pimm.shared_memory.NumpySMAdapter(shape=frame.shape, dtype=frame.dtype)
+            adapter.array[:] = frame
+            obs[logical] = adapter
+        return obs
+
+    def privileged(self, raw_obs: dict[str, Any]) -> dict[str, Any]:
+        return {'sim_state': raw_obs['sim_state']}
+
+    def terminal(self, result: dict[str, Any]) -> dict[str, Any] | None:
+        return {'eval.success': True} if result['done'] else None
```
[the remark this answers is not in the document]
vertix: Fixed in 2ed6516. `RemoteEnvControlSystem.reset` now emits `done.emit({})` synchronously, so any terminal the previous trial left on the wire (the env can reach `done` while the proxy free-runs between trials) is cleared before the harness samples it, rather than only on the later reset-publish turn.

Worth noting this is the proxy's terminal lifecycle, exposed for the first time by LIBERO: it's the first adapter with a *live* `terminal` — native `stack_cubes` returns `None` (scored downstream), so nothing previously emitted a mid-run success to leak. The env-server proxy tests (frame-0 publish + free-run) still pass with the synchronous clear.

positronic/simulator/libero/launcher.py
vertix: No need to make it a control system if we don't use it for anything pimm-related. это натягивание совы на глобус

positronic/simulator/libero/adapter.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def reset_token(self, seed: int | None) -> Any:
+        self._players = {}
+        self._held = {}
+        self._grip = 0.0
+        return seed if seed is not None else 0  # LIBERO selects a concrete init-state index
+
+    def action(self, commands: dict[str, pimm.Message], now_ns: int) -> dict[str, Any]:
+        for name, msg in commands.items():
+            player = self._players.setdefault(name, roboarm_command.TrajectoryPlayer())
```
vertix: make self._players defaultdict instead

positronic/simulator/libero/adapter.py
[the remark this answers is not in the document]
vertix: Don't be lazy and please validate it here. There's no value in shipping non-working code, even if someone will consume it later.
Also, use the same approach for other comments. The code must be fully functional

positronic/simulator/libero/validate.py
vertix: I want to keep it, but I want it to be run on CI. If you confirm it is part of CI, you can resolve this comment

positronic/simulator/libero/launcher.py
vertix: Done in adaf823 — `LiberoServer` is gone. `RemoteEnvControlSystem` now owns the env server's lifetime itself, via a `serve` context manager handed to it at construction: it enters it when it connects (first reset) and exits it after closing the socket in its `run` finally. For LIBERO that context manager is `serve_libero` (spawns the 3.10 subprocess, terminates it on exit); the mujoco env-server tests pass `nullcontext((host, port))` since their server already runs in-thread.

So the only control system is the proxy, and the subprocess teardown rides the proxy's World lifecycle — no more control-system-just-to-be-torn-down. Bonus: the proxy now closes the socket *before* the server dies (cleanup runs in reverse of acquisition), so shutdown ordering is structural rather than the earlier best-effort tolerant close.

positronic/simulator/libero/adapter.py
vertix: Done in adaf823 — `self._players` is a `defaultdict(roboarm_command.TrajectoryPlayer)` now (in both `LiberoAdapter` and the reference `StackCubesAdapter` it mirrors), so the `action` loop is just `player = self._players[name]`.

positronic/simulator/libero/adapter.py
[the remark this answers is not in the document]
vertix: Done and validated on-box in adaf823 — agreed, no point shipping it unverified.

I drove the gripper open→closed on a LIBERO box and read the finger qpos: open ≈ `[0.0397, -0.0397]`, closed ≈ `[0.0005, -0.0005]`. So `np.sum` cancels to ~0 exactly as you flagged, while the aperture `|q0 - q1|` runs 0.079 (open) → 0.001 (closed).

The normalization now lives **server-side** in `LiberoEnv._observe`, mirroring the native sim's `_current_grip` and the reference adapter's pass-through: `closure = 1 - |q0 - q1| / aperture_max`, clipped to [0, 1], where `aperture_max` comes from the model's two finger joint ranges (0.04 + 0.04 = 0.08) — no client-side magic constant. The adapter just passes `grip` through, and the raw `gripper_qpos` no longer crosses the wire.

`validate.py` got a `_check_grip` (open < 0.05, closed > 0.95) that now runs in the `libero-e2e` lane, so the mapping is CI-enforced. On-box it settles open ≈ 0.007, closed ≈ 0.987. I also confirmed the camera flip the same way — LIBERO's own video path flips `obs[cam][::-1]`, so our flip matches; dropped that TODO too.

positronic/simulator/libero/validate.py
vertix: Done in adaf823 — `validate.py` now runs in the `libero-e2e` CI lane (a step after the demo replay, reusing the LIBERO checkout the replay clones, same `MUJOCO_GL=osmesa`). It covers the OSC_POSE / JOINT_POSITION / JOINT_VELOCITY `set_goal` inverses, FK identity, the IK round-trip, and now a gripper-normalization check — the algebra the demo replay can't reach (it only drives OSC pose). Resolving per your note.

positronic/cfg/eval/sim/libero.py
vertix: On `instruction`: I verified it on-box and found a real mismatch — LIBERO's actual task language for `libero_spatial`/0 is `pick up the black bowl between the plate and the ramekin and place it on the plate` (lowercase, no period), but the override hardcoded a title-cased string with a trailing period. A LIBERO-trained policy conditions on the exact string, so that mattered. Fixed `spatial` to LIBERO's verbatim language in adaf823 and dropped the TODO.

The server already reports the same string in frame-0 meta (`task`), so sourcing `instruction` straight from the env — overrides carry only `suite`/`task_id`, no hardcoded language to drift — is the clean way to add more suites. That's a `Task`/harness change (instruction becomes env-sourced rather than a static field), so I left it for your call. Leaving this open: do you want (a) specific `(suite, task_id)` overrides added now for the tasks you'll run, or (b) the meta-sourced instruction so overrides stop hardcoding it?

positronic/policy/harness.py
```diff
@@ -282,6 +282,8 @@ def _build_obs(self, clock: pimm.Clock) -> dict[str, Any] | None:
         inputs['wall_time_ns'] = time.time_ns()
         inputs['obs_time_ns'] = clock.now_ns()
         inputs.update(self.context)
+        if self._task is not None:
+            inputs['task'] = self._task.instruction  # the eval owns the instruction; it overrides any context key
```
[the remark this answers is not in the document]
vertix: Fixed in 2d4ece6. You're right — resolving the instruction live moved its read to the inference/finalize sites, which dropped `task` from `self.context` (the dict passed to `new_session` and to the completion counter), collapsing `SampledPolicy` task-grouped sampling/counting and opening sessions without the task fixed-task evals used to provide.

Restored it by resetting the scene **before** opening the session in `_begin_episode`, then snapshotting the resolved instruction into `self.context` as before — so `task` reaches both the session and the counter, while a remote env can still report its task on reset (the proxy caches it synchronously in `reset()`, so `task.instruction` is available immediately after). The per-step/finalize injections are reverted since `task` flows through the context again. Added `test_task_instruction_reaches_session_context_after_reset`, which uses an instruction resolvable only after reset, to pin this.

positronic/simulator/env_server/proxy.py
```diff
@@ -71,14 +97,16 @@ def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[p
                     # The reset is this turn's step: publish the env's frame-0 (no step) and clear the prior
                     # terminal, so the recorder samples it before any step advances the env.
                     self._reset_pending = False
-                    self.robot_meta.emit(self._frame['meta'])
+                    self.robot_meta.emit(self._meta)
```
vertix: `self._meta` contains non-robot related data, like the prompt. I think the robot meta should be a separate field in the meta that servers sends. 

Enforce it for libero and for existing stack cubes test server

positronic/simulator/libero/env.py
```diff
@@ -0,0 +1,303 @@
...
+        if self._env is not None:
+            self._env.close()
+            self._env = None
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description='Serve one LIBERO task over the env-server protocol.')
+    parser.add_argument('--host', default='localhost')
+    parser.add_argument('--port', type=int, required=True)
+    parser.add_argument('--suite', required=True, help='LIBERO task suite, e.g. libero_spatial')
```
vertix: Why anything but --host and --port are required? I was under impression that server reads it from reset token

positronic/simulator/env_server/proxy.py
vertix: Fixed in dd0c096. Split the env-server reset payload into two fields: `meta` (scene/task identity — the prompt + scene ids, read by the `Task` for its live instruction) and a new `robot_meta` (robot model identity), each routed to its own sink. The proxy now emits `frame['robot_meta']` on the `robot_meta` port (→ episode static) instead of the prompt-bearing `meta`.

Enforced in both servers: the native stack-cubes server sends its model (`sim.robot_meta` — URDF/joints) as `robot_meta` with an empty scene `meta`; LIBERO sends `{suite, task_id, task}` as `meta` with an empty `robot_meta` — the 3.10 server can't import positronic to emit the Panda model, so the eval keeps supplying it via `static_meta`/`bundled_panda_model`. The degenerate test env got the field too.

Nothing is lost by no longer routing the prompt through `robot_meta`: the instruction is already recorded via `context['task']` (`harness._begin_episode`), and `suite`/`task_id` have no downstream consumer. env_server + harness tests pass.

positronic/simulator/libero/env.py
vertix: You're right — fixed in 7f5326d. The env-server protocol already specifies token-driven construction ("Construct (cached) + re-randomize from a token… cache keyed by the token's structural part; no build phase"), and `env.py` was the deviation: `--suite`/`--task-id`/`--camera-resolution`/`--control-mode` were CLI args fixed at process start.

Now the subprocess takes only `--host`/`--port`. The task spec rides the reset token: `LiberoAdapter` carries it and ships `{suite, task_id, camera_resolution, control_mode, seed}` in every token; `LiberoEnv()` builds the task on the first reset and caches it, rebuilding only if the spec changes — so one server can serve any task without a restart. `serve_libero()` lost the scene args; the e2e and `validate.py` build the token directly.

Validated on-box: the demo-replay e2e (reaches success through the socket) and `validate.py` (all three control modes, grip/FK/IK) both pass.

positronic/simulator/env_server/server.py
```diff
@@ -99,11 +104,21 @@ def _handle(self, connection: ServerConnection) -> None:
...
+        # ``serve`` would spawn a thread per connection; instead the accept loop runs here and calls the
+        # handshake + ``_handle`` inline, so the env runs on this thread (the subprocess main thread — macOS GLFW
+        # must init there). The loop skips bare TCP probes (which never handshake) and exits once the one client
+        # has been served.
         with serve(self._handle, self._host, self._port, max_size=None) as server:
             self._server = server
-            server.serve_forever()
+            while not self._served:
+                try:
+                    sock, addr = server.socket.accept()
```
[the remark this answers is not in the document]
vertix: Fixed in b395522. You're right that the custom accept loop's blocking `socket.accept()` wasn't woken by closing the listening socket from `shutdown`'s thread, so a caller shutting down before any client connected leaked the serving thread and closed the env while it was still alive.

The loop now sets a 0.5s timeout on the listening socket and checks a `_shutdown` flag each iteration, so teardown is observed within one poll even with no client connecting; the accepted socket is put back to blocking for the long-lived session. Verified the no-client-then-shutdown path joins the thread cleanly (no leak), and the existing serve/shutdown tests still pass.

positronic/simulator/libero/env.py
```diff
@@ -0,0 +1,308 @@
...
+from robosuite.utils.transform_utils import get_pose_error, make_pose, mat2quat, quat2axisangle
+from server import EnvProtocol, EnvServer
+
+# LIBERO's package __init__ prompts on stdin for a dataset path when ``~/.libero/config.yaml`` is absent, which
+# would hang this unattended server on a clean machine. Answer its first-run prompt with the default ('n') so it
+# writes the config from LIBERO's own package-relative paths, non-interactively.
+builtins.input = lambda *_args, **_kwargs: 'n'
+
+from libero.libero import benchmark, get_libero_path  # noqa: E402 -- must follow the stdin shim above
+from libero.libero.envs import OffScreenRenderEnv  # noqa: E402
```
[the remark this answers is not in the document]
vertix: Declining — verified false positive for this stack (Python 3.10.17, pinned LIBERO checkout). Two independent reasons the `collections.Iterable` line never executes:

1. **It's shadowed dead code.** `bddl_base_domain.py` does `from ...envs.utils import *` (line 17) then `from ...envs.regions import *` (line 20), so the `MultiRegionRandomSampler` it instantiates for fixtures (line 604) resolves to `regions.base_region_sampler.MultiRegionRandomSampler`, not the `utils.py` one flagged here. Confirmed at runtime: `bddl_base_domain.MultiRegionRandomSampler.__module__ == 'libero.libero.envs.regions.base_region_sampler'`. That class's `_sample_quat` (base_region_sampler.py:106) already uses the correct `collections.abc.Iterable`; `utils.py:105` is never reached.

2. **The rotation sampler isn't called on our reset path anyway.** Every reset restores a saved full state via `set_init_state` (seed path → recorded init-state; replay path → the demo's state), so placement comes from the state, not live sampling. Instrumenting `utils.MultiRegionRandomSampler._sample_quat` across the cabinet/drawer fixture tasks `libero_90/0`, `/6`, `/7` showed **0 calls**.

Empirically, with `collections.Iterable` confirmed absent (`hasattr(collections, 'Iterable') == False` after importing robosuite + libero), `_env.reset()` succeeds for `libero_spatial/0` (both seed and exact-state) and all three `libero_90` cabinet/drawer fixtures; the committed `libero-e2e` lane is green on the current commit. A shim would guard a line that can't execute in this stack, so I'm not adding one. Leaving this open for your call.

PR #454, opened 2026-06-25
Generalize DreamZero eval wrap into `video_context_wrappers`

vertix:
## Summary
- Collapse DreamZero's frame-stacked eval-`wrap` into one reusable, parameterized config
  `video_context_wrappers` (`ErrorRecovery() | frame_stack | ChunkedSchedule()`) in
  `positronic/cfg/wrappers.py`, expressed in two knobs: `history_frames` (oldest look-back, in frames)
  and `stride`.
- `dreamzero_wrappers` (in `positronic/vendors/dreamzero/codecs.py`) is now a one-line `.override` of it
  — no structural twin; its `--wrap=@positronic.vendors.dreamzero.codecs.dreamzero_wrappers` CLI path is
  unchanged. The schedule matches the upstream *method* (frames at `stride` back from the current frame,
  oldest **pinned to the window start**), reproducing DreamZero's trained offsets
  `(-23/15, -16/15, -8/15, 0.0)` byte-exactly — the deliberate `-23` (not `-24`) emerges from the pin.
- New context-window schedules need no committed code — they are CLI overrides of the generic config.

### Default wrap is now owned by the config layer
- `cfg.wrappers.default_wrappers` is the **single definition** of the default eval wrap
  (`ErrorRecovery() | ChunkedSchedule()`). `positronic/policy/wrappers.py` no longer carries a
  `default_wrappers` constant.
- `Harness` and `main` are mechanism-only: `Harness(wrap=...)` defaults to `None` (runs the raw policy)
  and `main` requires `wrap`. The eval entry points — `run` and inference's `run_cfg` — supply the
  default from `cfg.wrappers.default_wrappers`. Production behaviour is unchanged (eval always goes
  through those entry points).
- Because the default lives in `cfg.wrappers`, configuronic relative resolution works, so a custom
  schedule can be selected with a relative ref:
  `--wrap=.video_context_wrappers --wrap.history_frames=24 --wrap.stride=1` (the full
  `@positronic.cfg.wrappers.video_context_wrappers` form also works).

## Test plan
- `dreamzero_wrappers` instantiates to `offsets_sec == (-23/15, -16/15, -8/15, 0.0)` (byte-exact);
  the generic config + relative ref resolve to the 25-frame dense schedule
  (`tuple((i-24)/15 for i in range(25))`) via both `run.override(...)` and the full configuronic CLI.
- `run`'s default `wrap` instantiates to the no-frame-stack `ErrorRecovery | ChunkedSchedule` pipeline.
- Updated the tests that relied on `Harness`/`main` auto-wrapping to pass `wrap` explicitly; full affected
  suites green: `positronic/policy/tests/`, `vendors/dreamzero/tests/`, `tests/test_inference_integration.py`,
  `simulator/env_server/tests/test_remote_env.py` (136 passed).
- `ruff check` + `ruff format --check` clean on all changed files.

### DIFF
--- a/positronic/cli/eval/run.py
+++ b/positronic/cli/eval/run.py
@@ -143,15 +143,7 @@ def main(
 @cfn.config(
     eval=placeholder, policy=policy_cfg.placeholder, trial_count=1, show_gui=False, wrap=wrappers_cfg.default_wrappers
 )
-def run(
-    eval: Eval,
-    policy,
-    trial_count,
-    show_gui,
-    output_dir=None,
-    inference_latency=False,
-    wrap=wrappers_cfg.default_wrappers,
-):
+def run(eval: Eval, policy, trial_count, show_gui, output_dir=None, inference_latency=False, *, wrap):
     """Run a selected eval (embodiment + task) through the shared inference harness."""
     # The trial plan: one RUN context per trial, consumed by the self-driving Harness. Per-trial seeds
     # are known upfront — ``--eval.seed`` + trial index, or an independent random draw per trial when
--- a/positronic/cli/eval/run.py
+++ b/positronic/cli/eval/run.py
@@ -18,7 +18,7 @@ from positronic.dataset.local_dataset import LocalDatasetWriter, load_all_datase
 from positronic.eval import Embodiment, Eval, Task
 from positronic.gui.dpg import DearpyguiUi
 from positronic.policy.base import SampledPolicy
-from positronic.policy.harness import Harness, default_wrappers
+from positronic.policy.harness import Harness
 
 logger = logging.getLogger(__name__)
 
@@ -69,7 +69,8 @@ def main(
     task: Task | None = None,
     trials: list[dict] | None = None,
     show_gui: bool = False,
-    wrap=default_wrappers,
+    *,
+    wrap,
 ):
     """Run inference for an embodiment, real or simulated.
 
--- a/positronic/cli/eval/run.py
+++ b/positronic/cli/eval/run.py
@@ -10,6 +10,7 @@ import pos3
 
 import pimm
 import positronic.cfg.policy as policy_cfg
+import positronic.cfg.wrappers as wrappers_cfg
 from positronic import utils, wire
 from positronic.cfg.eval import placeholder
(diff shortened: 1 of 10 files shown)


positronic/cli/eval/run.py
```diff
@@ -138,8 +140,18 @@ def main(
...
+    eval=placeholder, policy=policy_cfg.placeholder, trial_count=1, show_gui=False, wrap=wrappers_cfg.default_wrappers
+)
+def run(
+    eval: Eval,
+    policy,
+    trial_count,
+    show_gui,
+    output_dir=None,
+    inference_latency=False,
+    wrap=wrappers_cfg.default_wrappers,
```
vertix: What's the point in duplicating default? Also, it is wrong default here, because it's a Config, not Wrapper

vertix: Fixed in 01b257f — dropped the `wrap` default from `run`'s signature and made it keyword-only, so the `@cfn.config` decorator is the sole source of the default. The decorator binding is a `Config` that configuronic instantiates (the right place for it); the signature copy was both redundant and wrongly typed as you noted.

positronic/cli/eval/run.py
```diff
@@ -68,7 +69,8 @@ def main(
     task: Task | None = None,
     trials: list[dict] | None = None,
     show_gui: bool = False,
-    wrap=default_wrappers,
+    *,
+    wrap,
```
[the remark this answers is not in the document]
vertix: Intentional. This PR moves ownership of the default eval wrap to the config layer: `cfg.wrappers.default_wrappers` is the single definition, supplied by the eval entry points (`run` and inference's `run_cfg`). `Harness` and `main` are mechanism-only by design — `Harness(wrap=None)` runs the raw policy and `main` requires `wrap` — so selecting a wrapper is always an explicit, in-distribution decision rather than a hidden runtime default.

Re-adding a default wrapper to `main` would reintroduce the duplicated default this PR set out to remove; defaulting to `None` would silently drop scheduling/recovery. Every in-repo caller (the CLI configs and the integration tests) passes `wrap` explicitly. Leaving open for the maintainer's call.

PR #460, opened 2026-06-29
Add `web` operator console driver for remote evals

vertix:
## Summary
- Add `WebEvalUI`, a headless web operator console that streams the live eval cameras as MJPEG to a browser and turns Start/Finish/Abort button presses into harness directives — a drop-in directive source replacing the `eval_ui`/`keyboard` drivers, for running attended evals remotely over an SSH tunnel.
- Register it as the `web` eval driver: `positronic-inference web`.
- Broaden `Driver.gui` from `DearpyguiUi | None` to `pimm.ControlSystem | None` (the slot already holds `EvalUI`).

It reuses the existing server pattern from `drivers/webxr.py` (uvicorn in a daemon thread, turbojpeg encoding). Frames are JPEG-encoded in the control-system run loop so only immutable bytes cross to the server thread. No harness changes — remote teleop is a planned follow-up.

Binds `127.0.0.1` only; reach it from a laptop with `ssh -L 8080:localhost:8080 <robot-host>` then open `http://localhost:8080`.

## Test plan
- `ruff check` and `ruff format --check` pass on the three files.
- Import/registration smoke passes: `python -c "import positronic.gui.web; import positronic.inference; import positronic.cli.eval.run"` — the `web` driver config resolves without error.
- Not yet exercised live: end-to-end MJPEG streaming and the directive flow require a real robot run (`positronic-inference web`) with cameras, which has not been run.

### DIFF
--- a/positronic/gui/templates/eval_console.html
+++ b/positronic/gui/templates/eval_console.html
@@ -22,14 +22,10 @@
           <button class="btn btn-danger" data-action="abort">Abort</button>
         </div>
 
-        <div class="camera-grid">
-          {% for name in cameras %}
-          <figure class="camera">
-            <video data-camera="{{ name }}" autoplay muted playsinline></video>
-            <figcaption>{{ name }}</figcaption>
-          </figure>
-          {% endfor %}
-        </div>
+        <figure class="camera">
+          <video id="feed" autoplay muted playsinline></video>
+          <figcaption>{{ cameras | join(' · ') }}</figcaption>
+        </figure>
       </main>
     </div>
 
@@ -38,17 +34,15 @@
         button.addEventListener('click', () => fetch('/directive/' + button.dataset.action, { method: 'POST' }));
       }
 
-      for (const video of document.querySelectorAll('video[data-camera]')) {
-        startStream(video, video.dataset.camera);
-      }
+      startStream(document.getElementById('feed'));
 
-      function startStream(video, name) {
+      function startStream(video) {
         if (!('MediaSource' in window)) {
           console.error('MediaSource Extensions not supported in this browser');
           return;
         }
         const scheme = location.protocol === 'https:' ? 'wss' : 'ws';
-        const socket = new WebSocket(`${scheme}://${location.host}/video/${name}`);
+        const socket = new WebSocket(`${scheme}://${location.host}/video`);
         socket.binaryType = 'arraybuffer';
 
         let sourceBuffer = null;
--- a/positronic/gui/templates/eval_console.html
+++ b/positronic/gui/templates/eval_console.html
@@ -1,0 +1,85 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Positronic – Eval console</title>
+    <link rel="stylesheet" href="/static/styles.css" />
+    <link rel="stylesheet" href="/assets/eval.css" />
+</head>
+<body>
+    <div id="app-shell">
+      <main class="main">
+        <header class="page-header">
+          <div class="page-kicker">Attended eval</div>
+          <h1>Eval console</h1>
+          <p>Live camera feeds from the robot. Start, finish, or abort the current episode.</p>
+        </header>
+
+        <div class="controls-bar">
+          <button class="btn btn-primary" data-action="start">Start</button>
+          <button class="btn btn-ghost" data-action="finish">Finish</button>
+          <button class="btn btn-danger" data-action="abort">Abort</button>
+        </div>
+
+        <div class="camera-grid">
+          {% for name in cameras %}
+          <figure class="camera">
+            <video data-camera="{{ name }}" autoplay muted playsinline></video>
+            <figcaption>{{ name }}</figcaption>
+          </figure>
+          {% endfor %}
+        </div>
+      </main>
+    </div>
+
+    <script>
+      for (const button of document.querySelectorAll('button[data-action]')) {
+        button.addEventListener('click', () => fetch('/directive/' + button.dataset.action, { method: 'POST' }));
+      }
+
+      for (const video of document.querySelectorAll('video[data-camera]')) {
+        startStream(video, video.dataset.camera);
+      }
+
+      function startStream(video, name) {
+        if (!('MediaSource' in window)) {
+          console.error('MediaSource Extensions not supported in this browser');
+          return;
+        }
+        const scheme = location.protocol === 'https:' ? 'wss' : 'ws';
+        const socket = new WebSocket(`${scheme}://${location.host}/video/${name}`);
+        socket.binaryType = 'arraybuffer';
+
+        let sourceBuffer = null;
+        const pending = [];
+        const pump = () => {
+          if (sourceBuffer && !sourceBuffer.updating && pending.length > 0) {
+            sourceBuffer.appendBuffer(pending.shift());
+          }
+        };
+
+        socket.addEventListener('message', (event) => {
+          if (typeof event.data === 'string') {
+            const codec = `video/mp4; codecs="${event.data}"`;
+            if (!MediaSource.isTypeSupported(codec)) {
+              console.error('Unsupported codec', codec);
+              socket.close();
+              return;
+            }
+            const mediaSource = new MediaSource();
+            video.src = URL.createObjectURL(mediaSource);
+            mediaSource.addEventListener('sourceopen', () => {
+              sourceBuffer = mediaSource.addSourceBuffer(codec);
+              sourceBuffer.addEventListener('updateend', pump);
+              pump();
+            });
+          } else {
+            pending.push(new Uint8Array(event.data));
+            pump();
+          }
+        });
+      }
+    </script>
+</body>
+</html>
--- a/positronic/gui/templates/eval_console.html
+++ b/positronic/gui/templates/eval_console.html
@@ -65,6 +65,7 @@
             video.src = URL.createObjectURL(mediaSource);
             mediaSource.addEventListener('sourceopen', () => {
               sourceBuffer = mediaSource.addSourceBuffer(codec);
+              sourceBuffer.mode = 'sequence';
               sourceBuffer.addEventListener('updateend', pump);
               pump();
             });
--- a/positronic/gui/templates/eval_console.html
+++ b/positronic/gui/templates/eval_console.html
@@ -3,30 +3,20 @@
 <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <title>Positronic – Eval console</title>
+    <title>Eval console</title>
     <link rel="stylesheet" href="/static/styles.css" />
     <link rel="stylesheet" href="/assets/eval.css" />
 </head>
 <body>
-    <div id="app-shell">
-      <main class="main">
-        <header class="page-header">
-          <div class="page-kicker">Attended eval</div>
-          <h1>Eval console</h1>
-          <p>Live camera feeds from the robot. Start, finish, or abort the current episode.</p>
-        </header>
-
-        <div class="controls-bar">
-          <button class="btn btn-primary" data-action="start">Start</button>
-          <button class="btn btn-ghost" data-action="finish">Finish</button>
-          <button class="btn btn-danger" data-action="abort">Abort</button>
-        </div>
-
-        <figure class="camera">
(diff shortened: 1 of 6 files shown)


positronic/gui/web.py
```diff
@@ -0,0 +1,115 @@
...
+from fastapi.responses import HTMLResponse, StreamingResponse
+
+import pimm
+from positronic.policy.harness import Directive
+
+_LOG_CONFIG = {
+    'version': 1,
+    'disable_existing_loggers': False,
+    'handlers': {
+        'file': {'class': 'logging.FileHandler', 'formatter': 'default', 'filename': '/tmp/web_eval.log', 'mode': 'w'}
```
vertix: Remove this file logging, console is enough for now

positronic/gui/web.py
```diff
@@ -0,0 +1,115 @@
...
+        self._frames: dict[str, bytes] = {}
+
+    def run(self, should_stop: pimm.SignalReceiver, clock: pimm.Clock) -> Iterator[pimm.Sleep]:
+        app = FastAPI()
+        jpeg_encoder = turbojpeg.TurboJPEG()
+
+        @app.get('/')
+        async def index():
+            feeds = ''.join(f'<img src="/stream/{name}" alt="{name}">' for name in self.cameras)
+            return HTMLResponse(
```
vertix: We should have a proper templates, not this quick hacky implementation.
Overall the server should be visually nice and match our style (some examples can be seen in positronic_server.py and ../phail-website pages).

positronic/gui/web.py
```diff
@@ -0,0 +1,115 @@
...
+            f' >>> WEB eval console: http://127.0.0.1:{self.port}/  '
+            f'(tunnel with: ssh -L {self.port}:localhost:{self.port} <robot-host>) <<<'
+        )
+        print(banner)
+
+        while not should_stop.value:
+            for cam_name, camera in self.cameras.items():
+                cam_msg = camera.read()
+                if cam_msg.data is not None and cam_msg.updated:
+                    self._frames[cam_name] = jpeg_encoder.encode(cam_msg.data.array, quality=self.jpeg_quality)
```
vertix: We should have proper video streaming instead of just reusing jpeg encoder. I think modern python libs have out-of-the box support for video streaming.

positronic/gui/web.py
vertix: Done in 0397c02 — dropped the file `_LOG_CONFIG`; uvicorn now logs to the console.

positronic/gui/web.py
vertix: Done in 0397c02 — moved to a Jinja2 template (`gui/templates/eval_console.html`) reusing the shared `server/static/styles.css` design tokens (`.main`, `.page-header`, `.btn`/`.btn-primary`/`.btn-ghost`, `--danger`), with a small `gui/static/eval.css` for the camera grid. Did not inherit `base.html` since it hardcodes the dataset-browser `app.js`/nav.

positronic/gui/web.py
vertix: Done in 0397c02 — replaced MJPEG with H.264 streamed as fragmented MP4 over a WebSocket, played via MSE in a `<video>` tag (PyAV encoder, already a dep). Chose this over WebRTC/aiortc because WebRTC negotiates ICE/UDP and is fragile through a single forwarded TCP port, whereas fMP4-over-WS is tunnel-safe. Verified headless that the pipeline produces decodable H.264; the browser MSE codec string is derived from the avcC box and sent over the socket so it always matches. Live in-browser playback still needs a real robot run.

positronic/gui/web.py
```diff
@@ -0,0 +1,242 @@
...
+        self._subscribers: set[queue.Queue] = set()
+
+    def _open(self, height: int, width: int) -> None:
+        self._container = av.open(
+            self._buffer, mode='w', format='mp4', options={'movflags': 'frag_keyframe+empty_moov+default_base_moof'}
+        )
+        stream = self._container.add_stream(
+            'libx264',
+            rate=self._fps,
+            options={'preset': 'ultrafast', 'tune': 'zerolatency', 'profile': 'baseline', 'level': '3.1'},
```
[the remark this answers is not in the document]
vertix: Agreed, and verified: with `level=3.1` forced, a 1080p frame still declared level `0x1F` (3.1) — a non-conformant bitstream that lenient ffmpeg decodes but browsers/hardware reject. Fixed in 3696497 by dropping the forced level and letting x264 choose it. Re-tested: 720p declares `0x1F`, 1080p declares `0x28` (level 4.0), each conformant — and since the MSE codec string is derived from the avcC box, the client string adapts automatically (`avc1.42C028` for 1080p).

positronic/gui/web.py
```diff
@@ -0,0 +1,240 @@
...
+
+        config = uvicorn.Config(app, host='127.0.0.1', port=self.port)
+        server = uvicorn.Server(config)
+        server_thread = threading.Thread(target=server.run, daemon=True)
+        server_thread.start()
+
+        banner = '=' * 80
+        print(banner)
+        print(
+            f' >>> WEB eval console: http://127.0.0.1:{self.port}/  '
```
vertix: a) use the normal IP, like we use in webxr. 127.0.0.1 won't work.
b) don't we need https for video streaming to work?

positronic/gui/web.py
```diff
@@ -0,0 +1,240 @@
...
+        app = FastAPI()
+        app.mount('/static', StaticFiles(directory=_shared_static()), name='static')
+        app.mount('/assets', StaticFiles(directory=_pkg_path('static')), name='assets')
+
+        @app.get('/', response_class=HTMLResponse)
+        async def index(request: Request):
+            return templates.TemplateResponse(request, 'eval_console.html', {'cameras': list(self._streams)})
+
+        @app.websocket('/video/{name}')
+        async def video(websocket: WebSocket, name: str):
```
vertix: we should have one video stream that would include all cameras. If later we'd love to "switch off" some cameras, we should do it in the UI.

positronic/gui/web.py
vertix: a) Done in 996826e — binds `0.0.0.0` and advertises `utils.resolve_host_ip()` in the banner, like webxr.
b) Not needed: MediaSource and `ws://` are not secure-context-gated, so the H.264/MSE video plays over plain `http://<host-ip>`. webxr requires https only because the WebXR *device* API mandates a secure context. If we ever want https here we can mirror webxr's self-signed cert path, but it's not required for the video — leaving this open in case you want it anyway.

positronic/inference.py
```diff
@@ -60,6 +61,15 @@ def make(output_dir: Path | None) -> Driver:
     return make
 
 
+@cfn.config(port=8080)
+def web(port, task):
+    def make(output_dir: Path | None) -> Driver:
+        ui = WebEvalUI(task=task, port=port)
```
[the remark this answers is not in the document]
vertix: Agreed — fixed in 996826e. `WebEvalUI` no longer takes or stores a `Task`; the `web` driver builds `WebEvalUI(port=port)` and emits `Directive.RUN()` with no task. The harness sets `context["task"] = self._task.instruction` on RUN (harness.py:221-222), so the instruction is injected regardless and nothing unpicklable is captured into the spawned background process.

positronic/gui/web.py
vertix: Done in 996826e — all cameras are tiled into a single H.264 stream (one `<video>`, one `/video` WebSocket). Each frame is resized to a common height and laid out in a row so it encodes as one stream. Per-camera show/hide can be a UI-side concern later (select/crop regions of the combined frame).

positronic/gui/web.py
```diff
@@ -0,0 +1,259 @@
...
+    """
+
+    def __init__(self, port=8080, fps=30, keyframe_interval=15, bitrate=8_000_000):
+        self.port = port
+        self.fps = fps
+        self.keyframe_interval = keyframe_interval
+        self.bitrate = bitrate
+        self.cameras = pimm.ReceiverDict(self, default=None)
+        self.directive = pimm.ControlSystemEmitter(self)
+        self._stream = _CameraStream(fps, keyframe_interval, bitrate)
```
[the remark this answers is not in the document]
vertix: Confirmed and fixed in d93918a. Reproduced it: `pickle.dumps(WebEvalUI())` raised `cannot pickle '_thread.lock' object` (the combined-stream refactor had moved `_CameraStream`, which holds a `threading.Lock`, into `__init__`). The stream and frame buffer are now created as locals inside `run()` — after the spawn boundary — and the WebSocket handler closes over the local. `WebEvalUI` now pickles cleanly (bare and with camera receiver slots wired).

positronic/gui/web.py
```diff
@@ -0,0 +1,259 @@
...
+                case 'start':
+                    self.directive.emit(Directive.RUN(), clock.now_ns())
+                case 'finish':
+                    self.directive.emit(Directive.FINISH(), clock.now_ns())
+                case 'abort':
+                    self.directive.emit(Directive.ABORT(), clock.now_ns())
+                case _:
+                    raise HTTPException(status_code=404)
+
+        config = uvicorn.Config(app, host='0.0.0.0', port=self.port)
```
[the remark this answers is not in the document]
vertix: Acknowledged, but keeping `0.0.0.0` as a deliberate choice. This mirrors `drivers/webxr.py`, which already binds all interfaces and exposes unauthenticated robot control over the LAN — the console needs to be reachable on the host IP for the operator's remote-access path (binding `127.0.0.1` + SSH forwarding does not fit how this is accessed), and it runs on a trusted lab/robot network. The exposure concern is real but not specific to this PR; if we want to require auth it should be a project-wide change covering webxr too, which is out of scope here.

positronic/gui/templates/eval_console.html
```diff
@@ -0,0 +1,79 @@
...
+            const codec = `video/mp4; codecs="${event.data}"`;
+            if (!MediaSource.isTypeSupported(codec)) {
+              console.error('Unsupported codec', codec);
+              socket.close();
+              return;
+            }
+            const mediaSource = new MediaSource();
+            video.src = URL.createObjectURL(mediaSource);
+            mediaSource.addEventListener('sourceopen', () => {
+              sourceBuffer = mediaSource.addSourceBuffer(codec);
```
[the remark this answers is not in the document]
vertix: Good catch — fixed in 18d3804. The SourceBuffer is now set to `sequence` mode right after creation, so a client connecting mid-stream gets fragments re-timestamped from 0 and the `<video>` plays immediately instead of stalling before the first buffered range.

positronic/gui/web.py
```diff
@@ -0,0 +1,260 @@
...
+            except WebSocketDisconnect:
+                pass
+            finally:
+                stream.unsubscribe(subscriber)
+
+        @app.post('/directive/{action}')
+        async def directive(action: str):
+            match action:
+                case 'start':
+                    self.directive.emit(Directive.RUN(), clock.now_ns())
```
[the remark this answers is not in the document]
vertix: Agreed — fixed in 18d3804. The web driver now carries the instruction like the keyboard and EvalUI drivers: `WebEvalUI` takes a `task: str | None` and emits `Directive.RUN(task=self.task)`, and the `web` config threads `task` through (mirroring `keyboard`). This reverses the earlier removal: that was to dodge an unpicklable payload, but the actual spawn blocker was the `threading.Lock` (fixed in d93918a) — a string task is pickle-safe, verified `pickle.dumps` on both `WebEvalUI(task=...)` and `Directive.RUN(task=...)`. When the harness itself has a Task it still overrides with `self._task.instruction`, so the task-bearing path is unchanged.

positronic/gui/templates/eval_console.html
```diff
@@ -0,0 +1,80 @@
...
+        }
+        const scheme = location.protocol === 'https:' ? 'wss' : 'ws';
+        const socket = new WebSocket(`${scheme}://${location.host}/video`);
+        socket.binaryType = 'arraybuffer';
+
+        let sourceBuffer = null;
+        const pending = [];
+        const pump = () => {
+          if (sourceBuffer && !sourceBuffer.updating && pending.length > 0) {
+            sourceBuffer.appendBuffer(pending.shift());
```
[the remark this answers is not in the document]
vertix: Agreed — fixed in 5b54949. The pump now evicts buffered ranges more than `MAX_BUFFER_SECONDS` (10s) behind `video.currentTime` before appending, so the SourceBuffer holds a bounded live window instead of growing until it hits the MSE quota. The `updating` guard serializes the `remove` against appends (the remove fires `updateend`, which re-enters the pump to continue appending).

PR #462, opened 2026-06-29
Sweep a full eval suite from one `eval run`

vertix:
## Summary
One `positronic eval run` now sweeps a whole eval's trial space — no external bash loop. Builds on #459 (the LIBERO env-server, now in main).

- **`task_id` rides the reset token.** `Task.reset` takes the per-trial run context; the Harness hands it the whole context; `EnvAdapter.reset_token(context)` / `proxy.reset(context)`; `LiberoAdapter` drops `task_id` from `__init__`. `MujocoSim.reset(seed)` is left unchanged (it doubles as the env-server token decoder), so the native config adapts via a lambda. **Env-server transport unchanged** — the server already caches per `(suite, task_id, …)`.
- **Eval configs own the trial sweep.** `build_trials(seed, trial_count, task_ids)` builds the per-trial RUN contexts (task range × seed); `Eval` gains a `trials` field; `Task.seed` is removed. Sim, LIBERO, and real configs all build their trials.
- **`main(evals)`** loops evals (a World + self-driving Harness per eval) and owns the policy lifetime: warmup / `pos3.sync` / seed-counter run once, the sampler counter is shared across evals, and `policy.close()` runs once. The Harness no longer closes the policy (that nulled `RemotePolicy._client` and would kill a multi-eval sweep after the first). `run` overlays the CLI-owned `inference_latency`.

Usage: `--eval=.sim.libero.spatial` sweeps all 10 spatial tasks; `--eval.task_id=3 --eval.trial_count=5` pins task 3 over 5 seeds.

## Test plan
- `uv run --locked pytest --no-cov` over the affected suites — harness (42), policy dir (127), sim-integration + env-server (12) — all pass; repo-wide collection (610) imports clean.
- `ruff check` + `ruff format` clean.
- A live LIBERO full-suite sweep needs the 3.10 env-server subprocess + a served policy and was **not** run here.

### DIFF
--- a/positronic/cfg/eval/sim/libero.py
+++ b/positronic/cfg/eval/sim/libero.py
@@ -1,5 +1,6 @@
 import configuronic as cfn
 
+from positronic.cfg.eval import build_trials
 from positronic.dataset.serializers import Serializers
 from positronic.drivers.roboarm import command as roboarm_command
 from positronic.drivers.roboarm.models import bundled_panda_model
@@ -8,6 +9,10 @@ from positronic.simulator.env_server.proxy import RemoteEnvControlSystem
 from positronic.simulator.libero.adapter import LiberoAdapter
 from positronic.simulator.libero.launcher import serve_libero
 
+# Task count per LIBERO suite — the config expands an unbound ``task_id`` over ``range(num_tasks)``. positronic
+# cannot import LIBERO (it lives in the 3.10 env server), so the counts are pinned here against the pinned commit.
+_SUITE_NUM_TASKS = {'libero_spatial': 10, 'libero_object': 10, 'libero_goal': 10, 'libero_10': 10, 'libero_90': 90}
+
 
 @cfn.config(
     camera_dict={'image.agentview': 'agentview_image', 'image.wrist': 'eye_in_hand_image'},
@@ -15,8 +20,10 @@ from positronic.simulator.libero.launcher import serve_libero
     control_mode='ee',
     timeout=20.0,
     seed=None,
+    task_id=None,
+    trial_count=1,
 )
-def _libero_eval(suite, task_id, timeout, camera_dict, camera_resolution, control_mode, seed):
+def _libero_eval(suite, task_id, trial_count, timeout, camera_dict, camera_resolution, control_mode, seed):
     """A LIBERO eval: the embodiment proxies a remote LIBERO env, the task carries the scenario.
 
     A LIBERO *suite* is a set of related manipulation tasks; ``task_id`` selects one within it — a fixed scene and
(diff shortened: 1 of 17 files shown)


positronic/cli/eval/run.py
```diff
@@ -140,30 +116,59 @@ def main(
...
+            for ev in evals:
+                _run_world(
+                    policy, ev.embodiment, ev.task, ev.trials, None, output_dir, show_gui, on_complete, wrap=wrap
+                )
+    finally:
+        policy.close()
+
+
+@cfn.config(eval=placeholder, policy=policy_cfg.placeholder, show_gui=False, wrap=wrappers_cfg.default_wrappers)
+def run(eval: Eval, policy, show_gui, output_dir=None, inference_latency=False, *, wrap):
```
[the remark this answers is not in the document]
vertix: Fixed in 524faad — took the "update the callers/docs" half of your suggestion, not the top-level alias.

The trial sweep now belongs to the eval config (the design thesis of this PR), so the new path is `--eval.trial_count` — consistent with `--eval.timeout` already used in the same sentence of `docs/inference.md`. Updated all three documented commands (`docs/inference.md:72`, `positronic/vendors/dreamzero/README.md:47` and `:152`).

Declined the "preserve a top-level `--trial_count` alias on `run`" option deliberately: re-adding it would resurrect the exact `run.trial_count` pathway this PR removes, leaving two owners for the value (a silent bridge). Note the default is unchanged — old `run` defaulted to `trial_count=1`, same as the native eval config now, so no behavior regression, only the flag path moved.

positronic/cfg/eval/sim/libero.py
```diff
@@ -69,15 +76,17 @@ def _libero_eval(suite, task_id, timeout, camera_dict, camera_resolution, contro
...
+    return Eval(embodiment, task, build_trials(seed, trial_count, task_ids))
 
 
-# Each suite binds only its LIBERO ``suite``; ``--task_id`` selects the trial. The instruction is not pinned —
-# the task reads it live from the env's reset meta. The task count per suite is in ``_libero_eval``'s docstring.
+# Each suite binds only its LIBERO ``suite``; an unbound ``task_id`` sweeps the whole suite, ``--eval.task_id``
+# pins one. The instruction is not pinned — the task reads it live from the env's reset meta.
 
 # libero_spatial — 10 tasks
 spatial = _libero_eval.override(suite='libero_spatial')
```
vertix, line 104: Please add another `all` config that runs all tasks.

positronic/simulator/libero/adapter.py
```diff
@@ -42,19 +42,14 @@ def __init__(
         camera_dict: dict[str, str],
         *,
         suite: str,
```
vertix: I'm thinking, we moved the `task_id`, but is there any particular reason why we should not move `suite`, `control_mode`, `settle_steps` and even may be `camera_resolution` to reset token?

All these things are per-session, and as long as the env supports it, they should be dynamic. What do you think?

positronic/cfg/eval/sim/libero.py
vertix: Added in ba642a6: `all` sweeps the four-suite LIBERO benchmark (spatial, object, goal, libero_10 — 40 tasks) in one run, the set papers report as "LIBERO average". `libero_90` stays out since it's the LIBERO-100 pretraining pool, not a standard eval target; it remains available as its own config.

Since the scene spec now rides each trial's context (see the adapter thread), `all` is just `_libero_eval.override(suite=[...])` — one embodiment, one env server, 40 trials. `--eval.task_id` composes with it (pins that task in every suite), and `--eval.trial_count`/`--eval.seed` sweep per scene as before.

positronic/simulator/libero/adapter.py
vertix: Agreed — done in ba642a6. All four (`suite`, `control_mode`, `camera_resolution`, `settle_steps`) moved out of the adapter constructor into the per-trial context; `reset_token` now reads them from the trial. The wire format didn't change at all — these fields already rode every reset token and the server already keys its env cache on them; only the client-side binding was static. The adapter is now scene-free (it keeps only `camera_dict`, which is client-side obs wiring, not scene).

`build_trials(task_ids)` generalized to `build_trials(scenes)` — a list of scene-spec context bases swept over the seed set — which is what makes the multi-suite `all` config a plain override.

This is decision D5 of #464 (the eval-spine design), which parks the broader restructuring this points at.

PR #501, opened 2026-07-23
Add i2rt YAM embodiment support

### LINKED ISSUE #396: Grouped time-series panes have unreadable/ambiguous labels after prefix stripping
The `positronic-server` episode viewer groups numeric signals by prefix before the first `.` (`_group_signals_by_prefix`) and renders each group as a pane in the bottom Rerun grid (`_build_blueprint`, `dataset_utils.py:182-187`). Grouping is good, but the labelling after stripping the prefix makes the panes hard to read:

- **Single-signal groups** are named after the *group prefix only*, discarding the leaf. E.g. an episode whose only `robot_commands.*` signal is `robot_commands.pose` shows a pane labelled just `robot_commands` — you cannot tell it is the pose stream.
- **Multi-signal groups** become `rrb.Tabs` whose tab names are the leaf with the prefix stripped (`pose`, `reset`, `q`, `dq`, …). Out of context these short tokens are ambiguous and easy to misread.

This is why the same logical signal looks different across episodes (e.g. a `robot_commands` pane reads differently depending on whether a stray `robot_commands.reset` was recorded — see #395).

**Suggested direction (open to discussion):** keep the prefix grouping but make labels self-describing — e.g. use the full signal name (`robot_commands.pose`) as the view/tab name, or show `prefix › leaf`, so a pane is unambiguous regardless of how many signals share the prefix.

RikHeijdens:
# Add i2rt YAM bimanual station: real-arm driver, cameras, teleop, and recording

Brings up the i2rt **YAM** bimanual station end to end — real-arm driver, MuJoCo
assets, station cameras, teleop entry, and the recording/viewer plumbing needed to
collect episodes on it.

### YAM arm support
- **Vendored MJCF assets** from I2RT under `positronic/assets/mujoco/i2rt_yam/`
  (`yam.xml`, `scene.xml`, meshes, upstream LICENSE/CHANGELOG/README).
- **`yam` extra** in `pyproject.toml` with a git-pinned `i2rt` dependency (`uv.lock` updated).
- **Real-arm driver** `positronic/drivers/roboarm/yam.py` for the i2rt arms, with a
  fake-mode smoke path for CI/no-hardware runs.
- **Configs, embodiments, and teleop entry**: `yam_bimanual` embodiment and roboarm
  config wired into `positronic/cfg/`.
- **basedpyright baseline** for `yam.py` — cheaply fixable errors fixed (optional-import
  narrowing, dtype, fake-mode narrowing); the remainder are the same classes already
  grandfathered for `franka.py`/`so101`.

### Cameras
- **Mono ZED X One** support added to `SLCamera` (`positronic/drivers/camera/zed.py`).
- **YAM station cameras** wired into `yam_bimanual` and recorded at **SVGA / 30 fps**.

### Recording & video
- The YAM recording config opts into
  a real-time x264 preset via the new `video_options` plumbing. Per-writer encoding threads stay.
- Video signals are **encoded on a per-writer thread** to keep recording off the hot path.
- `mounts` config parameter recorded into dataset `static_meta`.
- 
### GUI / viewer
- Episode viewer now **labels bimanual channels** (`positronic/gui/eval.py`,
  `positronic/server/dataset_utils.py`).

## Testing
- `uv run --locked pytest --no-cov`
- Fake-mode smoke exercises the YAM driver without hardware.

## Notes
- Adds the `yam` optional extra; install with the `yam` group to pull `i2rt`.
- Closes #396


vertix:
@codex review

positronic/cfg/embodiment.py
```diff
@@ -4,8 +4,10 @@
 import positronic.cfg.hardware.camera
 import positronic.cfg.hardware.gripper
 import positronic.cfg.hardware.roboarm
+from positronic import geom
 from positronic.dataset.serializers import Serializers
 from positronic.drivers.roboarm import command as roboarm_command
+from positronic.drivers.roboarm import yam as yam_driver
```
vertix: This will trigger import of the hardware libraries in `yam.py`. That one has protection, but nethertheless it might cost an extra glitch during start. Pls move these imports inside the functions.

There's a hook in this repo that complains when things are imported in methods (Claude used to do it quite often), but in case of configs this is a desired behavior.

positronic/drivers/roboarm/yam.py
```diff
@@ -0,0 +1,390 @@
...
+
+# Duplicated from the yam-bimanual branch's simulator/mujoco/yam.py so driver and sim agree on frames without
+# a code dependency. TODO: converge on one module when the YAM sim lands on this branch.
+_JOINT_NAMES = ('joint1', 'joint2', 'joint3', 'joint4', 'joint5', 'joint6')
+# The menagerie "home" keyframe: arm folded up and back, out of the workspace.
+HOME_Q = np.array([0.0, 1.047, 1.047, 0.0, 0.0, 0.0])
+# Sim world-frame mount geometry the training data uses: table top z=0.30, arms at (0.30, ±0.305) facing +x,
+# base plate lifted 0.011 above the tabletop.
+YAM_MOUNTS = {'left': (0.30, 0.305), 'right': (0.30, -0.305)}
+TABLE_Z = 0.30
```
vertix: These constants (YAM_MOUNTS, table_z, yam_mount_lift) belong to the config parameters (embodiment.py). They are not inherit to the driver itself, as different users of this robot may have different positions and configs.

pyproject.toml
```diff
@@ -112,7 +115,14 @@ dev = [
 [tool.uv]
 # lerobot_0_3_3 extra pulls in lerobot==0.3.3 which declares rerun-sdk<0.23,
 # but our usage of lerobot doesn't need rerun — override to unblock resolution.
-override-dependencies = ["rerun-sdk==0.30.0"]
+# i2rt (yam extra) exact-pins its environment (numpy==2.2.6, click<8.2.0, qpsolvers==4.7.1);
+# relax those to floors so an optional hardware extra doesn't downgrade the whole project.
+override-dependencies = [
```
vertix: Two asks:

1. With these overrides the YAM station runs numpy 2.3.5, but i2rt pins 2.2.6. Please
   confirm you've actually driven the arms on 2.3.5 — not just that it installs.

2. Please drop the three new overrides and instead add `{extra = "yam"}` to the
   `conflicts` block. I tested it on your branch: it resolves cleanly and nothing
   outside the yam extra changes. The catch is that `yam` can then no longer be
   installed together with `lerobot` or `molmoact2` — if the YAM machine needs those
   locally, tell me and we'll keep the overrides instead.

positronic/cfg/embodiment.py
```diff
@@ -43,6 +45,81 @@ def droid(robot_arm, gripper, cameras):
...
+    """Real bimanual i2rt YAM — channel names and static_meta match ``mujoco_yam_bimanual`` by convention.
+
+    Per-arm channels are the flat names the whole stack shares: ``robot_state.{side}`` expands into
+    ``robot_state.{side}.q/.dq/.ee_pose`` on record, commands are ``robot_command.{side}`` +
+    ``target_grip.{side}``, and static_meta lists ``arms=[left, right]``. Per-side ``base_pose`` mounts
+    each arm at the sim world-frame mount, so real ``ee_pose`` lands in the training world frame.
+    """
+    mount_z = yam_driver.TABLE_Z + yam_driver.YAM_MOUNT_LIFT
+    arms = {
+        side: yam_driver.Robot(channel, base_pose=geom.Transform3D([*yam_driver.YAM_MOUNTS[side], mount_z]))
```
vertix: I recommend adding these arguments (mount_z, left and right positions) into the meta. The meta got saved into the datasets, that will let you to have everything recorded, when you decide to move arms here or there.

positronic/dataset/video.py
```diff
@@ -30,12 +38,15 @@ def __init__(
...
             fps: Frame rate for encoding (default: 30)
+            codec_options: Encoder options. The default trades ~2x bitrate for ~2.5x faster encoding
+                (x264's medium preset caps a live recorder well below camera rate on weak CPUs).
         """
         self.video_path = video_path
         self.frames_index_path = frames_index_path
         self.codec = codec
         self.gop_size = gop_size
         self.fps = fps
+        self.codec_options = {'preset': 'ultrafast', 'tune': 'zerolatency'} if codec_options is None else codec_options
```
vertix: The `ultrafast`/`zerolatency` default needs to go back to `None`. As written it doubles the bitrate of every dataset anyone records, and `codec_options` can't actually be set by anyone — `local_dataset.py:150` builds the writer with paths only, so the changed default is the only route to the new behavior. Please pass it down instead, so only the YAM station opts in:

1. `video.py` — keep the `codec_options` param, revert the default to `None`.

2. `local_dataset.py`, `DiskEpisodeWriter.__init__` — add a keyword-only `video_options: dict[str, str] | None = None`, store it, and use it at line 150:
```python
VideoSignalWriter(video_path, frames_index, codec_options=self._video_options)
```

3. `local_dataset.py`, `LocalDatasetWriter.__init__` — add the same param and store it. `new_episode()` doesn't need a new argument, it just forwards the stored value:
```python
writer = DiskEpisodeWriter(ep_dir, created_ts_ns=created_ts_ns, uid=uid, video_options=self._video_options)
```

4. `data_collection.py`, `main()` — add a `video_options` param and pass it: `LocalDatasetWriter(output_dir, video_options=video_options)`.

Then `yamcfg` sets `video_options={'preset': 'ultrafast', 'tune': 'zerolatency'}` and the YAM station gets its fast encoding while everyone else keeps today's quality.

Threading itself looks good, keep it.

positronic/server/dataset_utils.py
```diff
@@ -193,7 +193,7 @@ def _ts_view(name: str, sig: str) -> rrb.TimeSeriesView:
         if len(sigs) == 1:
             view = _ts_view(group_name, sigs[0])
         else:
-            view = rrb.Tabs(*[_ts_view(sig[len(group_name) + 1 :], sig) for sig in sigs], name=group_name)
+            view = rrb.Tabs(*[_ts_view(sig, sig) for sig in sigs], name=group_name)
```
vertix: This is related to #396 — can you please:

1. Link it and mark that this PR closes it.
2. Finish the other half. The issue has two symptoms; this fixes tab names in multi-signal groups. The other is in the branch right above: a group with a single signal is named after the prefix and the leaf is dropped, so a pane showing only `robot_command.pose` is labeled just `robot_command` and you can't tell which stream it is.

Both arguments to `_ts_view` are the same value at every call site now, so they can collapse:

```python
def _ts_view(sig: str) -> rrb.TimeSeriesView:
    return rrb.TimeSeriesView(
        name=sig,
        origin=f'/signals/{sig}',
        plot_legend=rrb.PlotLegend(visible=signals.dims.get(sig, 1) > 1),
        axis_y=rrb.ScalarAxis(zoom_lock=True),
    )

series_views = []
for group_name, sigs in _group_signals_by_prefix(signals):
    if len(sigs) == 1:
        view = _ts_view(sigs[0])
    else:
        view = rrb.Tabs(*[_ts_view(sig) for sig in sigs], name=group_name)
    series_views.append(view)
```

positronic/server/dataset_utils.py
```diff
@@ -231,15 +231,20 @@ def _ts_view(name: str, sig: str) -> rrb.TimeSeriesView:
 
 
 def _setup_series_names(signals: EpisodeSignals, ep: Episode) -> None:
-    joint_signal = ep.static.get('joint_signal')
+    joint_set = set(ep.static.get('joint_signals', ()))
```
vertix: `joint_signal` is read in three places and each does its own `ep.static.get`. Only this one knows about the plural key — `_log_numeric_signals` (which picks the arrays kept for the 3D panel) and `_log_urdf_robot` still look for the singular one. Can you route all three through one accessor?

```python
def _joint_signals(ep: Episode) -> set[str]:
    names = set(ep.static.get('joint_signals', ()))
    if ep.static.get('joint_signal'):
        names.add(ep.static['joint_signal'])
    return names
```

Keep the fallback — published datasets have `joint_signal` baked into their `static.json`, so it can't go yet. We opened #511 to remove it for real (needs changes to the read-time transforms and the public release configs) and will take that once this lands. A `TODO(#511)` on the fallback line would be good.

positronic/data_collection.py
```diff
@@ -355,6 +355,18 @@ def so101cfg(robot_arm, **kwargs):
...
+@cfn.config(
+    robot_arm=positronic.cfg.hardware.roboarm.yam,
+    webxr=positronic.cfg.webxr.oculus,
+    sound=positronic.cfg.sound.sound,
+    operator_position=OperatorPosition.BACK,
+    cameras={},
+)
+def yamcfg(robot_arm, **kwargs):
+    """Runs data collection on a real i2rt YAM arm (the arm driver carries the gripper)."""
+    main(robot_arm=robot_arm, gripper=robot_arm, **kwargs)
```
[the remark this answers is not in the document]
vertix: Pre-existing and not YAM-specific — `so101cfg` passes `gripper=robot_arm` the same way today. The conflation is in `main`: `bg_cs` treats "which ports" and "which processes" as one list, and `World.start` spawns one process per entry.

Fixed both in #513. Your config is fine as written; rebase onto main, otherwise `positronic-data-collection yam` raises at startup.

positronic/cfg/embodiment.py
```diff
@@ -43,6 +45,81 @@ def droid(robot_arm, gripper, cameras):
...
+        **{
+            f'robot_command.{s}': Command(arm.commands, roboarm_command.Reset(), Serializers.robot_command)
+            for s, arm in arms.items()
+        },
+        **{f'target_grip.{s}': Command(arm.target_grip, 0.0, None) for s, arm in arms.items()},
+    }
+    static_meta = {
+        'joint_signals': [f'robot_state.{s}.q' for s in arms],
+        'pose_signals': [f'robot_state.{s}.ee_pose' for s in arms],
+        'command_pose_signals': [f'robot_command.{s}.pose' for s in arms],
```
vertix: Signals in `pose_signals` get two things: axis labels on the plots, and a trail in the 3D view. Signals in `command_pose_signals` only get the labels — so on bimanual you won't see the commanded pose in 3D at all. Single-arm does show it, as the orange trail.

Just merge them:

```python
'pose_signals': [f'robot_state.{s}.ee_pose' for s in arms] + [f'robot_command.{s}.pose' for s in arms],
```

Labels still work, and the `| set(ep.static.get('command_pose_signals', ()))` line in `dataset_utils.py` goes away.

positronic/drivers/roboarm/yam.py
```diff
@@ -0,0 +1,390 @@
...
+from collections.abc import Callable, Iterator
+from typing import Any
+
+import mujoco as mj
+import numpy as np
+
+try:
+    from i2rt.robots.get_robot import get_yam_robot
+    from i2rt.robots.utils import GripperType
+except ImportError:
```
vertix: Pls drop the sentinel and raise with the install hint instead, the way `franka.py` does:

```python
except ImportError as e:
    raise ImportError('YAM support is not installed. Install the yam extra:\n  uv sync --locked --extra yam\n') from e
```

That also removes the `None` check in `_connect`. As written, a broken i2rt install — any one of its 19 deps missing — reports "not installed" and sends you to reinstall what you already have.

The `--fake` smoke is the only thing that needs this module to import without i2rt, and it's a bring-up script rather than a unit test — same category as franka's `__main__`, which needs `positronic-franka` installed to run at all. So requiring `--extra yam` for yours is consistent. I checked it installs and imports cleanly on macOS arm64.

#517 tracks pulling the MJCF kinematics out of the driver. That part has no i2rt dependency and deserves a real test. Not for this PR.

positronic/gui/dpg.py
```diff
@@ -1,6 +1,11 @@
 from collections.abc import Iterator
 
-import dearpygui.dearpygui as dpg
+try:
```
vertix: This is fixed on main now — #516 defers the import into the call sites and markers the dependency out on ARM:

```toml
"dearpygui; platform_machine != 'aarch64'",
```

Pls drop this hunk and the matching one in `gui/eval.py`. After a rebase, ARM Linux installs and imports without either guard, and `dpg.py` keeps its plain top-level import instead of a module-level `dpg = None` that every `dpg.<x>` call in the file would then depend on.

You'll need to rebase onto main and re-run `uv lock` — #516 touched `uv.lock` too.

`positronic.gui.dpg_ui()` is the constructor if you need to build the UI from new code.

positronic/drivers/roboarm/yam.py
```diff
@@ -0,0 +1,390 @@
...
+    get_yam_robot = GripperType = None
+
+import pimm
+from positronic import geom
+from positronic.utils import package_assets_path
+
+from . import RobotStatus, State, command
+from .ik import qpos_from_site_pose
+
+# Duplicated from the yam-bimanual branch's simulator/mujoco/yam.py so driver and sim agree on frames without
```
vertix: Three things here point at code that isn't on main, and a fresh reader can't resolve any of them.

1. This comment names a branch, and the TODO is conditioned on that branch landing. Comments should read
   for someone who only has main — state the invariant, and hang the TODO off an issue:

```python
# The driver solves FK/IK itself, so its joint order and frames must match the YAM sim's.
# TODO(#NNN): share one module between driver and sim.
```

2. `yam_bimanual`'s docstring (`cfg/embodiment.py:81`) says the channel names "match `mujoco_yam_bimanual`".
   That config doesn't exist — not in this PR, not on main.

3. `assets/mujoco/i2rt_yam/scene.xml` is added but nothing loads it — `_Kinematics` reads `yam.xml` directly.
   Fine to keep if you're shipping the vendored package whole, or if the sim half needs it. Just say which.

Is the sim PR right behind this one? If it lands soon, 1 and 2 mostly sort themselves out and all I'd want
here is the issue number in the TODO.

positronic/drivers/roboarm/yam.py
```diff
@@ -0,0 +1,390 @@
...
+        :param sim: Run against i2rt's own MuJoCo sim instead of hardware.
+        :param connect: ``(channel, sim) -> i2rt Robot`` factory; the fake-mode smoke injects ``_FakeYam``.
+        """
+        self._channel = channel
+        self._home_joints = np.asarray(home_joints if home_joints is not None else HOME_Q, dtype=np.float64)
+        self._base_pose = base_pose if base_pose is not None else geom.Transform3D.identity
+        self._sim = sim
+        self._connect = connect
+
+        self.commands: pimm.SignalReceiver[command.CommandType] = pimm.ControlSystemReceiver(self, default=None)
```
vertix: These two annotations are wrong, and it changes what the `Reset()` branch below has to do.

The channels also carry **trajectories**. `policy/harness.py:310` emits a list of `(ts_ns, value)` waypoints on
every command channel, `target_grip` included:

```python
traj = [(int(a['timestamp'] * 1e9), a[name]) for a in actions if name in a]
emitter.emit(traj)
```

Teleop still emits bare values (`data_collection.py:162,164`), and `TrajectoryPlayer.set` accepts both — which
is why the single-value annotations look plausible. They came from `so101/driver.py:58-59`, where we never
updated them after trajectories landed. Ours to fix, not yours, but pls don't carry them forward.

Two asks:

1. Drop both annotations for now — #518 tracks getting them right across all the drivers.

```python
self.commands = pimm.ControlSystemReceiver(self, default=None)
self.target_grip = pimm.ControlSystemReceiver(self, default=0.0)
```

2. Clear both players on `Reset()`. Since chunks really are in flight, `grip_target = 0.0` gets overwritten by
   the next due waypoint of the old chunk on the very next tick, and the arm resumes stale waypoints away from
   home. Both `franka.py:262` (on error recovery) and `simulator/mujoco/sim.py:229-230` (on `reset()`) already
   do this:

```python
case command.Reset():
    player.set([])
    grip_player.set([])
    q_target = self._reset(arm, kin, robot_state)
    grip_target = 0.0
```

vertix (changes requested):

positronic/drivers/roboarm/yam.py
vertix: One correction to the snippet above: use `uv run`, not `uv sync`. `main` has since settled on this form (#521,
closing #520 — franka had the same problem):

```python
except ImportError as e:
    raise ImportError(
        'YAM support is not installed. Re-run with the yam extra:\n  uv run --locked --extra yam ...\n'
    ) from e
```

`uv sync --locked --extra yam` only works from inside a checkout, so it swaps one misleading hint for another —
which is what franka's original `pip install` hint did. `uv run --extra` names the command you were already
running, and works anywhere.

Same change also retires the `pip install "positronic[yam]"` in the comment on line 24; after #521 there is no
`pip install` hint left in the Python sources.

positronic/cfg/embodiment.py
RikHeijdens: Moved both imports into `yam_bimanual` — the only user. `cfg.hardware.roboarm.yam` already
deferred the driver import. Since the driver now raises at import without i2rt (your other
comment), the deferral is also what keeps the base install importable.

positronic/drivers/roboarm/yam.py
RikHeijdens: Moved to `cfg/embodiment.py`: added a `mounts` config parameter to `yam_bimanual`
(side → world-frame `[x, y, z]` of the arm base, defaulting to the sim scene geometry), and the
driver keeps only `HOME_Q` and the joint names. `mounts` is also recorded in `static_meta` per
your other comment, so each dataset carries the mount survey it was recorded with.

pyproject.toml
RikHeijdens: Done — the three overrides are gone and `{extra = "yam"}` joined the `conflicts` block;
re-locked. Verified the base resolution is unchanged vs main, and the yam fork now resolves
i2rt's exact pins (numpy 2.2.6, click 8.1.8, qpsolvers 4.7.1) — so going forward the station
runs the numpy i2rt was tested with, and the 2.3.5 question retires itself.

We drove the arms using numpy 2.2.6 when testing the changes after addressing PR feedback. We previously used 2.3.5 which seemed to work fine as well. We didn't need the `lerobot` or `molmoact2` extras on the control station.

positronic/cfg/embodiment.py
RikHeijdens: Done — see https://github.com/Positronic-Robotics/positronic/pull/501#discussion_r3656330975; `static_meta['mounts']` now records the per-side mount positions.

positronic/dataset/video.py
RikHeijdens: Done exactly as laid out: `codec_options` default back to `None`, `video_options` threaded
`main()` → `LocalDatasetWriter` → `DiskEpisodeWriter` → `VideoSignalWriter`, and only `yamcfg`
opts into `{'preset': 'ultrafast', 'tune': 'zerolatency'}`. Kept threading.

positronic/server/dataset_utils.py
RikHeijdens: Done — `_ts_view` collapsed to a single argument and single-signal groups are named by the full
signal path, so both #396 symptoms are fixed. PR description updated to close #396.

positronic/server/dataset_utils.py
RikHeijdens: Done — `_joint_signals(ep)` merges the plural key with the singular fallback (fallback line
carries `TODO(#511)`), and all three readers (`_setup_series_names`, `_log_numeric_signals`,
`_log_urdf_robot`) go through it.

positronic/cfg/embodiment.py
RikHeijdens: Done — commanded poses ride in `pose_signals`; the `command_pose_signals` union in
`dataset_utils.py` is gone.

positronic/drivers/roboarm/yam.py
RikHeijdens: Done with the `uv run --locked --extra yam ...` wording — sentinel, the `None` check in
`_connect`, and the `pip install` comment are all gone. The `--fake` smoke now needs the yam
extra, same category as franka's `__main__`. Verified:
`uv run --locked --extra yam python -m positronic.drivers.roboarm.yam --fake` passes on the
new lockfile (i2rt fork env, numpy 2.2.6).

positronic/gui/dpg.py
RikHeijdens: Rebased onto main and dropped the guard commit — `gui/dpg.py` and `gui/eval.py` are identical
to main again, and `uv lock` was re-run on top of #516's lockfile.

positronic/drivers/roboarm/yam.py
RikHeijdens: 1. The constants comment now states the invariant and hangs the TODO off #517.
2. `yam_bimanual`'s docstring no longer references `mujoco_yam_bimanual`; it states the channel
   naming convention on its own terms.
3. The upstream MJCF package is vendored whole (`scene.xml`, meshes, LICENSE/CHANGELOG/README);
   the driver loads only `yam.xml`. Stated in the module docstring now.

We have sim support working on a local branch, but I think it should receive some additional attention to validate the implementation/camera parameters. The sim support reads `scene.xml`.

positronic/drivers/roboarm/yam.py
RikHeijdens: The rebase mooted the annotation question in a nicer way: after #523 (closing #518) the ports
really do carry `Trajectory[...]`, so yam now mirrors `franka.py`/`so101` — annotated
`Trajectory[CommandType]` / `Trajectory[float]` with `default=[]` — instead of dropping the
annotations. `Reset()` clears both players, and the `__main__` smoke emits
`[(now_ns, value)]` trajectories to match the new wire shape.

vertix:
@codex review

positronic/cfg/embodiment.py
```diff
@@ -43,6 +43,87 @@ def droid(robot_arm, gripper, cameras):
...
+        **{
+            f'robot_command.{s}': Command(arm.commands, roboarm_command.Reset(), Serializers.robot_command)
+            for s, arm in arms.items()
+        },
+        **{f'target_grip.{s}': Command(arm.target_grip, 0.0, None) for s, arm in arms.items()},
+    }
+    static_meta = {
+        'joint_signals': [f'robot_state.{s}.q' for s in arms],
+        'pose_signals': [f'robot_state.{s}.ee_pose' for s in arms] + [f'robot_command.{s}.pose' for s in arms],
+        'arms': list(arms),
```
vertix: `arms` is redundant with `mounts` directly below it, and nothing reads it — I grepped for
`static['arms']` / `static.get('arms')` and there are no consumers anywhere in the repo. The side names are
already recoverable twice over from this same dict: from `mounts`, and from the `joint_signals` prefixes.

The two can also drift, because they come from different places — `arms` from the hardcoded `('left', 'right')`
pair on line 95, `mounts` from a free config parameter:

```python
mounts={'left': [...], 'right': [...], 'third': [...]}   # 'third' recorded, but no arm is built for it
mounts={'left': [...]}                                   # KeyError at line 94
```

So the recorded `arms` list doesn't pin down what was actually recorded any better than `mounts` does.

Either fix is fine by me, your pick:

1. Drop `'arms'` entirely — `mounts` already names the sides. The docstring above (`static_meta lists
   arms=[left, right]`) needs updating to match.
2. Keep the key but derive it from the same source, so the two can't disagree:

```python
'arms': list(mounts),
```

Small either way — don't hold the PR on it if you'd rather do it as a follow-up.

vertix (dismissed):

vertix (approved):

PR #543, opened 2026-07-30
Record real frame timing in the video container

vertix:
`VideoSignalWriter` stored an ordinal frame index in `pts` — a field that means time — against a fixed declared rate (`fps=100`). Reader and writer agreed through that rate, so reads inside the library were correct. Everything outside reads `pts` as time: a player, a browser, ffmpeg, the viewer — so the defect was invisible from here. On a pre-existing recording (`sweep_jp` episode 5, 159 frames) the frames arrive 66 ms apart (15.15 fps) spanning 10.428 s, while the container declares 100 fps and 1.590 s — 6.56x too fast, over a sixth of the real duration. True timing lived only in `frames.parquet`.

## What it stores now

Each frame is muxed with a microsecond presentation timestamp taken from its `ts_ns`, relative to the first frame, on a `1/1000000` time base — microseconds, not nanoseconds, because some MP4 boxes carry a 32-bit timescale and duration.

The frames index gains a `pts` column, making the index the single owner of the mapping between frame position and container time: the reader looks it up instead of re-deriving it from a rate convention both sides assume. Random access seeks to a frame's recorded `pts` and recovers a decoded frame's index by bisecting that array — decode order cannot be counted, since a seek restarts at the preceding keyframe.

Two consequences:

- `fps` leaves `VideoSignalWriter` — it declared a fact the writer is not given and cannot source, while the writer already receives the truth on every frame. The `rate=` passed to `add_stream` survives only as a nominal encoder hint for bitrate and GOP heuristics, named and commented as such.
- Presentation timestamps are guarded monotonic: `append` enforces strictly-increasing `ts_ns`, but two frames inside one microsecond round to the same tick, which the muxer rejects. Such a frame is bumped to `previous + 1`, and the index records what was actually muxed.

Seeking needed a related correction: seeks are decode-timestamp based while these targets are presentation timestamps, so with B-frames a keyframe's decode timestamp can precede the target while its presentation timestamp follows it, leaving the stream past the wanted frame. The seek now steps back until decoding resumes at or before the target — latent before, and it applies on both paths.

This changes the on-disk container for newly written data: new `.mp4` files carry a microsecond time base and real timestamps instead of a `1/12800` base and ordinal ticks, and their `frames.parquet` carries the extra `pts` column. Existing recordings are untouched.

## Pre-existing exports stay readable

An index with no `pts` column falls back to exactly the previous behaviour — `ticks_per_frame` from the declared stream rate, `index = pts // ticks_per_frame` — the convention those files were written with. Checked against the recordings the live viewer serves: old and new readers return bit-identical frames and timestamps across 8 pre-existing videos, read sequentially, in reverse and in random order — 0 mismatches. A test writes a legacy-format fixture explicitly rather than via the current writer, pinning the fallback against the real old format.

## Measured

Re-recording that episode's frames through `LocalDatasetWriter`:

| | declared rate | container duration | vs recorded span |
|---|---|---|---|
| before | 100.00 fps | 1.590 s | 6.56x too fast |
| after | 15.20 fps | 10.461 s | 1.00x |

Recorded span is 10.428 s over 159 frames. Per-frame timing error after the change is 0 ns, max and mean, and index `pts` equal container `pts` exactly; the residual 33 ms is the final frame's own duration, not drift.

Tests in `positronic/dataset/tests/test_video.py` cover the container span under jittery, non-uniform spacing including a gap no single rate could describe; index `pts` against what the container carries; random access with a seek forced on every read; time search inside the gap; the sub-microsecond collision; and the legacy fixture. The `basedpyright` baseline shrinks by 6 entries; no file grows.

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -3924,56 +3924,8 @@
             {
                 "code": "reportOptionalMemberAccess",
                 "range": {
-                    "startColumn": 21,
-                    "endColumn": 26,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportOptionalMemberAccess",
-                "range": {
-                    "startColumn": 21,
-                    "endColumn": 27,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportOptionalMemberAccess",
-                "range": {
-                    "startColumn": 21,
-                    "endColumn": 28,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportOptionalMemberAccess",
-                "range": {
-                    "startColumn": 21,
-                    "endColumn": 29,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportOptionalMemberAccess",
-                "range": {
                     "startColumn": 39,
                     "endColumn": 45,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportArgumentType",
-                "range": {
-                    "startColumn": 59,
-                    "endColumn": 81,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportArgumentType",
-                "range": {
-                    "startColumn": 59,
-                    "endColumn": 81,
                     "lineCount": 1
                 }
             },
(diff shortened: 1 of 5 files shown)


v-positronic:
Closing unmerged, on the driver's call.

The symptom this came out of — recorded video playing wrong in the dataset viewer — was caused by a wide, high-dimensional signal starving the video panel, and is resolved by filtering those signals out of the view. The container-timing rewrite is not needed for it, and the on-disk container change it carries is not worth taking on its own.

Reads through the dataset API were always correct: true per-frame timing lives in `frames.parquet`, which is what the library indexes through. What stays true is that a raw player reading an mp4 straight out of storage sees the declared rate rather than the recorded one.

The branch `video-real-pts` keeps the work if that ever needs picking up.

PR #545, opened 2026-07-30
Codify repo conventions as named rules in `CODE_RULES.md`

vertix:
## Summary

Codifies repo conventions as named rules, read from one place by the Codex reviewer, by agents, and by
humans.

- **`CODE_RULES.md`** — judgment rules a linter cannot make, each with an id, an imperative statement,
  and the safe path. Three to start, all three from review comments on #485:
  - `caller-in-name` — don't name things after where they are used
  - `hardcoded-keys` — don't write data field names as literals inside the code
  - `overspecific` — don't bake assumptions about the caller into an interface; generalise until it
    simplifies, and stop there
- **`AGENTS.md`** — points the Codex reviewer at `CODE_RULES.md` and fixes the report format
  (`Rule {name} violated:`). Deliberately contains no rule text.
- **`CLAUDE.md`** — points at `CODE_RULES.md` too, so the rules reach Claude while it writes code, not
  only when it reviews. Sessions load `CLAUDE.md`, not `AGENTS.md`, so without this the rules would have
  been review-only.
- **`/add-rule`** — authors and recalibrates rules. A rule needs a real incident behind it, must not
  duplicate ruff or basedpyright, and has to survive a cold read by someone who was not in the
  conversation that produced it.
- **`/check-rules`** — checks a diff against the rules, one isolated agent per rule, no session context.
  Reports only; fixing stays a human decision.

Rules are waived at the site with `# rules-allow: <rule-id> — <reason>`, narrow like a `noqa`.

Also adds one `CLAUDE.md` rule that the rulebook's own prose failed: no empty intensifiers ("genuinely",
"truly", "load-bearing"). The offending words are removed from `CLAUDE.md` and `docs/architecture.md`.

## Open question this PR tests

Codex's documentation describes rules living under a `## Code Review Rules` heading in `AGENTS.md`, and
says nothing about following links to other files. This PR deliberately keeps `AGENTS.md` rule-free to
find out whether the link is followed. Partial evidence so far: Codex cited `CODE_RULES.md` by line
number while reviewing this branch — but that file was in the diff, so it is not yet proof. If Codex
never cites a rule id on a code diff, the fallback is a pre-commit hook that mirrors `CODE_RULES.md`
into `AGENTS.md`, with everything else unchanged.

## Not in this PR

- Seven further rule candidates: `optional-lies`, `silent-bridge`, `structural-twins`,
  `type-dispatch-hacks`, `impl-owns-default`, `comment-narrates-diff`, `speculative-shim`.
- Calibrating `overspecific` against the existing tree — it is the rule most likely to over-fire, and it
  has not been sampled yet.
- Resolving the overlap between `CODE_RULES.md` and the design-discipline section `CLAUDE.md` still
  carries; the two coexist for now.

## Test plan

Not verified — the mechanism can only be tested once Codex reviews a code diff with these files on the
branch. Markdown only; no code paths touched.

### DIFF
--- a/.claude/skills/add-rule/SKILL.md
+++ b/.claude/skills/add-rule/SKILL.md
@@ -68,9 +68,9 @@ the thing that went wrong.
 ## Step 4: Cold-read test
 
 The rule will be applied by readers with no access to this conversation. Test that directly with a
-subagent, which starts empty unless you fill it in. Give it the draft rule and two unlabelled snippets,
-one violating and one clean, and ask which violates it and why. Keep out which is which, the incident
-behind the rule, and what you hope it will say.
+subagent, which starts empty unless you fill it in. It needs the draft rule and two snippets, one
+violating and one clean, and must deliberately not know which is which, where they came from, or what
+you hope it will say. Ask which snippet violates the rule, and why.
 
 If it picks wrong, or its reasoning is not the reasoning you intended, the rule is unclear. Reword and
 repeat. Do not proceed on a rule that needed the conversation to be understood.
--- a/.claude/skills/add-rule/SKILL.md
+++ b/.claude/skills/add-rule/SKILL.md
@@ -1,12 +1,12 @@
 ---
 name: add-rule
-description: Author, calibrate, or retire a rule in CODE_RULES.md. Use when a review comment or a recurring code smell should become a durable repo rule, or when an existing rule is noisy, never fires, or keeps getting waived.
+description: Author or recalibrate a rule in CODE_RULES.md. Use when a review comment or a recurring code smell should become a durable repo rule, or when an existing rule produces findings people argue with.
 ---
 
 # Add Rule
 
-Every rule in `CODE_RULES.md` spends attention on every review, so each one earns its place or comes
-out. This skill writes one, calibrates it against the repo, or retires one.
+Every rule in `CODE_RULES.md` spends attention on every review, so each one earns its place. This skill
+writes one, or recalibrates one that has gone noisy.
 
 Nothing is written to `CODE_RULES.md` without the user's explicit approval of the final wording.
 
@@ -92,24 +92,3 @@ Search the tree for what it would flag, read a sample of the hits, and judge eac
 you agree with mean the rule is right and the repo has a backlog; hits you disagree with mean it is too
 broad and needs narrowing or dropping. Say how large the sample was, so a thin one cannot read as a
 clean sweep.
-
-## Retiring a rule
-
-Rules only accumulate unless something removes them, and a stale rule dilutes attention across the
-whole file. Retire one when:
-
-- it has not fired in review for a long stretch, and a scan of the tree finds nothing it would catch;
-- its waivers are out of proportion to its compliant uses. Count the markers, but treat the count as a
-  prompt to look, never as the evidence itself:
-
-  ```bash
-  grep -rn 'rules-allow: <rule-id>' --include='*.py' .
-  ```
-
-  Five waivers against five hundred compliant sites is a healthy rule with real exceptions; five against
-  eight is a rule describing the exception as the norm. Sample both sides before concluding. If the
-  waivers win, either the rule is wrong or the exception is the real rule; say which.
-- a linter can now enforce it. Move it to `pyproject.toml` and delete it here.
-
-Deleting a rule needs the same explicit approval as adding one. Say what it was, why it is going, and
-what replaces it if anything does.
--- a/.claude/skills/add-rule/SKILL.md
+++ b/.claude/skills/add-rule/SKILL.md
@@ -68,10 +68,9 @@ the thing that went wrong.
 ## Step 4: Cold-read test
 
 The rule will be applied by readers with no access to this conversation. Test that directly with a
-subagent, which starts empty unless you fill it in. Its prompt asks which snippet violates the rule and
-why, and carries exactly two pieces of context: the draft rule, and two unlabelled code snippets — one
-violating, one clean. Nothing else: not which is which, not the incident behind the rule, not what you
-hope it will say.
+subagent, which starts empty unless you fill it in. Give it the draft rule and two unlabelled snippets,
+one violating and one clean, and ask which violates it and why. Keep out which is which, the incident
+behind the rule, and what you hope it will say.
 
 If it picks wrong, or its reasoning is not the reasoning you intended, the rule is unclear. Reword and
 repeat. Do not proceed on a rule that needed the conversation to be understood.
--- a/.claude/skills/add-rule/SKILL.md
+++ b/.claude/skills/add-rule/SKILL.md
@@ -1,0 +1,120 @@
+---
+name: add-rule
+description: Author, calibrate, or retire a rule in CODE_RULES.md. Use when a review comment or a recurring code smell should become a durable repo rule, or when an existing rule is noisy, never fires, or keeps getting waived.
+---
+
+# Add Rule
+
+`CODE_RULES.md` is read by the Codex reviewer on every pull request, by Claude while implementing, and
+by `/check-rules`. Every rule in it spends attention on every review, so each one earns its place or
+comes out. This skill writes one, calibrates it against the repo, or retires one.
+
+Nothing is written to `CODE_RULES.md` without the user's explicit approval of the final wording.
+
+## What a rule is
+
+A judgment call no linter can make, stated as an instruction.
+
+A good rule is:
+
+- **Imperative** — "Don't X. Do Y instead." Not a description of the symptom; the reader should not have
+  to work out what to do.
+- **Explicit about the safe path**, including the exception where one exists.
+- **Checkable from a diff alone**, by someone who was not in the conversation that produced it.
+- **Durable** — describes behaviour, not the names of functions, files, or classes that exist today.
+- **Consequential** — worth someone stopping their work to fix.
+
+Not a rule:
+
+- Anything ruff or basedpyright can enforce. That belongs in `pyproject.toml`, where it runs
+  deterministically on every commit.
+- An aspiration — "keep interfaces clean", "name things well". Unfalsifiable, so it fires at random.
+- A symptom with no instruction attached.
+- Something so broad that most files violate it.
+
+## Step 1: Get the incident
+
+Ask what real code triggered this, and read it — the review comment, the commit, the function. A rule
+with no incident behind it is a preference; say so and stop.
+
+The incident is also the source of the bad/good example pair in Step 3, so capture it before it's lost.
+
+## Step 2: Rule out the cheaper homes
+
+Two checks, both of which kill the rule if they hit:
+
+```bash
+sed -n '/\[tool.ruff/,/^\[tool\.setuptools/p' pyproject.toml   # is it already a lint rule?
+grep -n '^### ' CODE_RULES.md                                  # does an existing rule cover it?
+```
+
+If ruff or basedpyright could catch it, enable the check there instead and stop. If an existing rule
+covers it, sharpen that rule rather than adding a second one — overlapping rules make findings
+ambiguous about which one was violated.
+
+## Step 3: Draft it
+
+```markdown
+### <rule-id>
+
+Don't <the thing>. <Do this instead>.
+
+Exception: <where it is legitimately fine>.
+```
+
+Then a bad/good code pair whenever the rule is about code shape rather than a naming or comment habit.
+Keep both halves to a few lines and take them from the Step 1 incident, stripped of anything
+repo-specific that will rot.
+
+`<rule-id>` is kebab-case and names the violation, not the remedy — it appears in review comments as
+`Rule <rule-id> violated:` and in waivers as `# rules-allow: <rule-id> — <reason>`, so it must read as
+the thing that went wrong.
+
+## Step 4: Cold-read test
+
+The rule will be applied by readers with no access to this conversation. Test that directly: hand a
+fresh subagent **only** the draft rule and two unlabelled code snippets — one violating, one clean —
+and ask which violates it and why.
+
+If it picks wrong, or its reasoning is not the reasoning you intended, the rule is unclear. Reword and
+repeat. Do not proceed on a rule that needed the conversation to be understood.
+
+## Step 5: Calibrate against the repo
+
+A rule that has never fired is a guess. Scan the existing tree — not just the recent diff — for code
+the rule would flag, then **read a sample of the hits** (five, or all of them if fewer) and judge each
+one yourself.
+
+The count sizes the sample; the sample decides. Report both, and the fork they imply:
+
+- **Hits you agree with, few.** Calibrated. Land it.
+- **Hits you agree with, many.** The rule is right and the repo has a backlog. Land it, and raise
+  separately whether to clean up now or let it bite on future changes — a rule that flags a lot of
+  existing code will make the next few reviews noisy, and the user should know that going in.
+- **Hits you disagree with.** The rule is too broad. Narrow it, or drop it.
+
+Report the sample honestly, including hits that make the rule look bad.
+
+## Step 6: Approve and append
+
+Show the final wording, the id, the hit count, and the sample. On explicit approval, append the rule to
+`CODE_RULES.md` in its section, then commit following the repo's conventions.
+
+## Retiring a rule
+
+Rules only accumulate unless something removes them, and a stale rule dilutes attention across the
+whole file. Retire one when:
+
+- it has not fired in review for a long stretch, and a scan of the tree finds nothing it would catch;
+- it is waived more often than it is obeyed — count the `# rules-allow:` markers carrying its id:
+
+  ```bash
+  grep -rn 'rules-allow: <rule-id>' --include='*.py' .
+  ```
+
+  A rule with several live waivers is describing an exception as if it were the norm. Either the rule
+  is wrong or the exception is the real rule; say which.
+- a linter can now enforce it. Move it to `pyproject.toml` and delete it here.
+
+Deleting a rule needs the same explicit approval as adding one. Say what it was, why it is going, and
+what replaces it if anything does.
--- a/.claude/skills/add-rule/SKILL.md
+++ b/.claude/skills/add-rule/SKILL.md
@@ -5,9 +5,8 @@ description: Author, calibrate, or retire a rule in CODE_RULES.md. Use when a re
 
 # Add Rule
 
-`CODE_RULES.md` is read by the Codex reviewer on every pull request, by Claude while implementing, and
-by `/check-rules`. Every rule in it spends attention on every review, so each one earns its place or
-comes out. This skill writes one, calibrates it against the repo, or retires one.
+Every rule in `CODE_RULES.md` spends attention on every review, so each one earns its place or comes
+out. This skill writes one, calibrates it against the repo, or retires one.
 
 Nothing is written to `CODE_RULES.md` without the user's explicit approval of the final wording.
 
@@ -95,10 +94,11 @@ The count sizes the sample; the sample decides. Report both, and the fork they i
 
 Report the sample honestly, including hits that make the rule look bad.
 
-## Step 6: Approve and append
+## Step 6: Approve and write
 
-Show the final wording, the id, the hit count, and the sample. On explicit approval, append the rule to
-`CODE_RULES.md` in its section, then commit following the repo's conventions.
+Show the final wording, the id, the hit count, and the sample. On explicit approval, write it to
+`CODE_RULES.md` — a new rule is appended, a revision of an existing id replaces that rule in place, so
+no id ever appears twice. Then commit following the repo's conventions.
 
 ## Retiring a rule
 
--- a/.claude/skills/add-rule/SKILL.md
+++ b/.claude/skills/add-rule/SKILL.md
@@ -40,12 +40,8 @@ The incident is also the source of the bad/good example pair in Step 3, so captu
 
 ## Step 2: Rule out the cheaper homes
 
-Two checks, both of which kill the rule if they hit:
-
-```bash
-sed -n '/\[tool.ruff/,/^\[tool\.setuptools/p' pyproject.toml   # is it already a lint rule?
-grep -n '^### ' CODE_RULES.md                                  # does an existing rule cover it?
-```
+Two checks, both of which kill the rule if they hit: whether ruff or basedpyright already covers it
+(`pyproject.toml`), and whether an existing rule does (`CODE_RULES.md`).
 
 If ruff or basedpyright could catch it, enable the check there instead and stop. If an existing rule
 covers it, sharpen that rule rather than adding a second one — overlapping rules make findings
--- a/.claude/skills/add-rule/SKILL.md
+++ b/.claude/skills/add-rule/SKILL.md
@@ -71,9 +71,10 @@ the thing that went wrong.
 
 ## Step 4: Cold-read test
 
-The rule will be applied by readers with no access to this conversation. Test that directly: hand a
-fresh subagent **only** the draft rule and two unlabelled code snippets — one violating, one clean —
-and ask which violates it and why.
+The rule will be applied by readers with no access to this conversation. Test that directly with a
+subagent, which starts empty unless you fill it in. Its prompt contains exactly two things: the draft
+rule, and two unlabelled code snippets — one violating, one clean. Not which is which, not the incident
+behind the rule, not what you hope it will say. Ask which snippet violates the rule, and why.
 
 If it picks wrong, or its reasoning is not the reasoning you intended, the rule is unclear. Reword and
 repeat. Do not proceed on a rule that needed the conversation to be understood.
--- a/.claude/skills/add-rule/SKILL.md
+++ b/.claude/skills/add-rule/SKILL.md
@@ -106,14 +106,16 @@ Rules only accumulate unless something removes them, and a stale rule dilutes at
 whole file. Retire one when:
 
 - it has not fired in review for a long stretch, and a scan of the tree finds nothing it would catch;
-- it is waived more often than it is obeyed — count the `# rules-allow:` markers carrying its id:
+- its waivers are out of proportion to its compliant uses. Count the markers, but treat the count as a
+  prompt to look, never as the evidence itself:
 
   ```bash
   grep -rn 'rules-allow: <rule-id>' --include='*.py' .
   ```
 
-  A rule with several live waivers is describing an exception as if it were the norm. Either the rule
-  is wrong or the exception is the real rule; say which.
+  Five waivers against five hundred compliant sites is a healthy rule with real exceptions; five against
+  eight is a rule describing the exception as the norm. Sample both sides before concluding. If the
+  waivers win, either the rule is wrong or the exception is the real rule; say which.
 - a linter can now enforce it. Move it to `pyproject.toml` and delete it here.
 
 Deleting a rule needs the same explicit approval as adding one. Say what it was, why it is going, and
--- a/.claude/skills/add-rule/SKILL.md
+++ b/.claude/skills/add-rule/SKILL.md
@@ -79,27 +79,19 @@ behind the rule, not what you hope it will say. Ask which snippet violates the r
 If it picks wrong, or its reasoning is not the reasoning you intended, the rule is unclear. Reword and
 repeat. Do not proceed on a rule that needed the conversation to be understood.
 
-## Step 5: Calibrate against the repo
+## Step 5: Approve and write
 
-A rule that has never fired is a guess. Scan the existing tree — not just the recent diff — for code
-the rule would flag, then **read a sample of the hits** (five, or all of them if fewer) and judge each
-one yourself.
-
-The count sizes the sample; the sample decides. Report both, and the fork they imply:
-
-- **Hits you agree with, few.** Calibrated. Land it.
-- **Hits you agree with, many.** The rule is right and the repo has a backlog. Land it, and raise
-  separately whether to clean up now or let it bite on future changes — a rule that flags a lot of
-  existing code will make the next few reviews noisy, and the user should know that going in.
-- **Hits you disagree with.** The rule is too broad. Narrow it, or drop it.
-
-Report the sample honestly, including hits that make the rule look bad.
-
-## Step 6: Approve and write
-
-Show the final wording, the id, the hit count, and the sample. On explicit approval, write it to
+Show the final wording and the id. On explicit approval, write it to
 `CODE_RULES.md` — a new rule is appended, a revision of an existing id replaces that rule in place, so
 no id ever appears twice. Then commit following the repo's conventions.
+
+## Calibrating a noisy rule
+
+Run this when a rule starts producing findings people argue with — not on every rule that lands.
+Search the tree for what it would flag, read a sample of the hits, and judge each one yourself. Hits
+you agree with mean the rule is right and the repo has a backlog; hits you disagree with mean it is too
+broad and needs narrowing or dropping. Say how large the sample was, so a thin one cannot read as a
+clean sweep.
 
 ## Retiring a rule
 
--- a/.claude/skills/add-rule/SKILL.md
+++ b/.claude/skills/add-rule/SKILL.md
@@ -68,9 +68,10 @@ the thing that went wrong.
 ## Step 4: Cold-read test
 
 The rule will be applied by readers with no access to this conversation. Test that directly with a
-subagent, which starts empty unless you fill it in. Its prompt contains exactly two things: the draft
-rule, and two unlabelled code snippets — one violating, one clean. Not which is which, not the incident
-behind the rule, not what you hope it will say. Ask which snippet violates the rule, and why.
+subagent, which starts empty unless you fill it in. Its prompt asks which snippet violates the rule and
+why, and carries exactly two pieces of context: the draft rule, and two unlabelled code snippets — one
+violating, one clean. Nothing else: not which is which, not the incident behind the rule, not what you
+hope it will say.
 
 If it picks wrong, or its reasoning is not the reasoning you intended, the rule is unclear. Reword and
 repeat. Do not proceed on a rule that needed the conversation to be understood.
--- a/.claude/skills/add-rule/SKILL.md
+++ b/.claude/skills/add-rule/SKILL.md
@@ -36,7 +36,7 @@ Not a rule:
 Ask what real code triggered this, and read it — the review comment, the commit, the function. A rule
 with no incident behind it is a preference; say so and stop.
 
-The incident is also the source of the bad/good example pair in Step 3, so capture it before it's lost.
+The incident is also where the rule's example comes from, so capture it before it's lost.
 
 ## Step 2: Rule out the cheaper homes
 
@@ -49,33 +49,17 @@ ambiguous about which one was violated.
 
 ## Step 3: Draft it
 
-```markdown
-### <rule-id>
-
-Don't <the thing>. <Do this instead>.
-
-Exception: <where it is legitimately fine>.
-```
-
-Then a bad/good code pair whenever the rule is about code shape rather than a naming or comment habit.
-Keep both halves to a few lines and take them from the Step 1 incident, stripped of anything
-repo-specific that will rot.
+Read the rules already in `CODE_RULES.md` and write one like them — same shape, same length, same
+plainness. Be concise and stay on the essence of the issue.
 
 `<rule-id>` is kebab-case and names the violation, not the remedy — it appears in review comments as
 `Rule <rule-id> violated:` and in waivers as `# rules-allow: <rule-id> — <reason>`, so it must read as
 the thing that went wrong.
 
-## Step 4: Cold-read test
+The rule will be applied by people who were not in the conversation that produced it. If it only makes
+sense to someone who was, reword it.
 
-The rule will be applied by readers with no access to this conversation. Test that directly with a
-subagent, which starts empty unless you fill it in. It needs the draft rule and two snippets, one
-violating and one clean, and must deliberately not know which is which, where they came from, or what
-you hope it will say. Ask which snippet violates the rule, and why.
-
-If it picks wrong, or its reasoning is not the reasoning you intended, the rule is unclear. Reword and
-repeat. Do not proceed on a rule that needed the conversation to be understood.
-
-## Step 5: Approve and write
+## Step 4: Approve and write
 
 Show the final wording and the id. On explicit approval, write it to
 `CODE_RULES.md` — a new rule is appended, a revision of an existing id replaces that rule in place, so
--- a/.claude/skills/check-rules/SKILL.md
+++ b/.claude/skills/check-rules/SKILL.md
@@ -33,13 +33,11 @@ Say which scope you used. A clean report over the wrong scope is worse than no r
 grep -n '^### ' CODE_RULES.md
 ```
 
-Spawn one agent per rule, in parallel. Each needs its rule and a way to tell which code changed, and
-must deliberately not know how that code came to be: it is a cold-headed reviewer answering one
-question, not a colleague who sat through the reasoning.
-
-One agent per rule, not one agent holding all of them — an agent given nine rules skims for nine and
-checks none. A narrow agent also fails visibly: reporting nothing is one rule cleanly checked, not
-nine rules half-checked.
+Spawn them in parallel, one rule each rather than one agent holding all of them — an agent given nine
+rules skims for nine and checks none, and reporting nothing then means nine rules half-checked instead
+of one cleanly checked. Each needs its rule and a way to tell which code changed, and must deliberately
+not know how that code came to be: a cold-headed reviewer answering one question, not a colleague who
+sat through the reasoning.
 
 Each agent returns two sections, either of which may be empty:
 
--- a/.claude/skills/check-rules/SKILL.md
+++ b/.claude/skills/check-rules/SKILL.md
@@ -33,10 +33,9 @@ Say which scope you used. A clean report over the wrong scope is worse than no r
 grep -n '^### ' CODE_RULES.md
 ```
 
-Spawn one agent per rule, in parallel. Give it the job, that one rule, and the change.
-
-Keep out anything that would tell it the answer: the other rules, what the change was for, why the code
-looks the way it does, what you expect it to find. It reads the repository itself for the rest.
+Spawn one agent per rule, in parallel. Each needs its rule and a way to tell which code changed, and
+must deliberately not know how that code came to be: it is a cold-headed reviewer answering one
+question, not a colleague who sat through the reasoning.
 
 One agent per rule, not one agent holding all of them — an agent given nine rules skims for nine and
 checks none. A narrow agent also fails visibly: reporting nothing is one rule cleanly checked, not
--- a/.claude/skills/check-rules/SKILL.md
+++ b/.claude/skills/check-rules/SKILL.md
@@ -77,6 +77,6 @@ Report only. Fixing is a separate decision and the user makes it.
 
 ## When a finding is wrong
 
-A false positive is information about the rule, not just about the code. If a rule produces findings
-you disagree with, take it to the `add-rule` skill and recalibrate or reword it there. Arguing with the
-same rule twice means the rule is broken.
+A false positive is information about the rule, not just about the code. Say so in the report and stop
+there: only a human changes `CODE_RULES.md`, through the `add-rule` skill and with their explicit
+approval. This skill never edits a rule, however wrong the rule looks from here.
--- a/.claude/skills/check-rules/SKILL.md
+++ b/.claude/skills/check-rules/SKILL.md
@@ -16,17 +16,18 @@ what stops it from seeing a violation.
 
 ```bash
 BASE=$(git merge-base HEAD main)
-git diff $BASE --stat                                      # base -> worktree
-git diff --cached $BASE --stat                             # base -> index, what a plain commit records
-git status --porcelain --untracked-files=all | grep '^??'  # untracked files, which no diff shows
+git diff $BASE HEAD --stat                                 # committed — what a push sends
+git diff $BASE --cached --stat                             # index — what a plain commit records
+git diff $BASE --stat                                      # worktree — what you see in the files
+git status --porcelain --untracked-files=all | grep '^??'  # untracked — what no diff shows
 ```
 
-Everything not yet on `main` is in scope: committed, staged, unstaged, and untracked. Pass the untracked
-files' contents alongside the diff — a brand-new file is where rules bite hardest and is the one thing
-`git diff` cannot show.
+Git keeps three snapshots — HEAD, the index, the worktree — and a violation can sit in any one of them
+while the others look clean: staged and then reverted, or committed and then fixed only in the index.
+Check all three, plus untracked files. That set is closed; there is no fourth place for code to hide.
 
-The two diffs are identical unless a file was staged and then edited again. When they differ, both are
-in scope: content can sit in the index, never appear in the worktree, and still be committed.
+In the ordinary case the three agree and it is one diff. Pass the untracked files' contents alongside
+it — a brand-new file is where rules bite hardest and is the one thing no diff shows.
 
 Narrow to the files the user named if they named any, and say which scope you used. A clean report over
 the wrong scope is worse than no report.
--- a/.claude/skills/check-rules/SKILL.md
+++ b/.claude/skills/check-rules/SKILL.md
@@ -8,9 +8,14 @@ description: Check the current changes against CODE_RULES.md and report violatio
 Checks a diff against `CODE_RULES.md`. Nothing else — not a code review, not a style pass, not a bug
 hunt. A finding that does not name a rule is out of scope, however good it is.
 
-The check runs in fresh subagents with no access to this conversation. That isolation is the point: the
-session that wrote the code knows why every line looks the way it does, and that knowledge is exactly
-what stops it from seeing a violation.
+Each rule is checked by a subagent that starts empty — no conversation history, no files this session
+has read, no other rule. That much the platform guarantees; the one thing that would undo it is a fork,
+which inherits the parent conversation, so never use one here.
+
+What the platform cannot guarantee is the prompt you write, and that is the only channel into the agent.
+You are the worst possible author for it: you know why every line looks the way it does, and that
+knowledge is precisely what stops you seeing the violation. Whatever of it you write into the prompt,
+the agent inherits.
 
 ## Step 1: Scope the diff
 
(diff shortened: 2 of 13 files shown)


CODE_RULES.md
```diff
@@ -0,0 +1,57 @@
+# Code Rules
+
+Judgment rules for this repository — the checks a linter cannot make. Read by the Codex reviewer on
```
vertix: It's not only for Codex or Claude, it's for humans and other agents, no need to state this

CODE_RULES.md
```diff
@@ -0,0 +1,57 @@
...
+pull requests, by Claude while implementing, and by the `check-rules` skill.
+
+Mechanical checks live in `pyproject.toml` (ruff, basedpyright), never here. A rule a linter could
```
vertix: Drop this line, noise. Overall try to be concise and focus on improtant

CODE_RULES.md
vertix: Fixed in 17edbdb9: dropped the reader list. The file no longer says who reads it.

CODE_RULES.md
vertix: Fixed in 17edbdb9: dropped it, along with two more lines of the same kind — the waiver section's "not that the rule is inconvenient" and "to retire a rule everywhere, delete it from this file". Preamble is 4 lines now.

The point it made survives where it is actionable: the `add-rule` skill's Step 2 refuses a rule that ruff or basedpyright could enforce, so the file doesn't need to assert it.

.claude/skills/add-rule/SKILL.md
```diff
@@ -0,0 +1,120 @@
+---
+name: add-rule
+description: Author, calibrate, or retire a rule in CODE_RULES.md. Use when a review comment or a recurring code smell should become a durable repo rule, or when an existing rule is noisy, never fires, or keeps getting waived.
+---
+
+# Add Rule
+
+`CODE_RULES.md` is read by the Codex reviewer on every pull request, by Claude while implementing, and
+by `/check-rules`. Every rule in it spends attention on every review, so each one earns its place or
```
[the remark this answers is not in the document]
vertix: Fixed in 9291c1de, though not the way you suggest: `/check-rules` now exists — it landed in 8ea3f412, one commit after the one you reviewed.

The line went anyway. It listed who reads `CODE_RULES.md`, and a human reviewer had already made the same objection about the identical sentence in `CODE_RULES.md` itself: the file is for humans and other agents too, so enumerating readers is both noise and always incomplete. The skill's intro now just says what the skill does.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,65 @@
...
+hunt. A finding that does not name a rule is out of scope, however good it is.
+
+The check runs in fresh subagents with no access to this conversation. That isolation is the point: the
+session that wrote the code knows why every line looks the way it does, and that knowledge is exactly
+what stops it from seeing a violation.
+
+## Step 1: Scope the diff
+
+```bash
+git diff --stat                                   # uncommitted work
```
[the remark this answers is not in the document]
vertix: Fixed in 9291c1de. Correct, and the same defect showed up from two more angles — untracked files and committed-but-unpushed work — so all three are fixed by one change rather than three patches.

Step 1 no longer picks one scope over another. It takes everything not yet on `main`:

```bash
BASE=$(git merge-base HEAD main)
git diff $BASE --stat                # committed, staged and unstaged, against main
git status --porcelain | grep '^??'  # untracked files, which no diff shows
```

`git diff $BASE` compares the merge base against the working tree, so staged and unstaged edits are both in, along with anything already committed on the branch.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,68 @@
...
```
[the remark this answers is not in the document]
vertix: Fixed in 9291c1de, together with the staged-edits and branch-commits findings — one scope change covers all three.

Your case was the most damaging of the three: a brand-new file is exactly where rules bite hardest, and it was the one thing the old command could not show at all. Step 1 now reads untracked paths from `git status --porcelain` and the skill says to pass their contents to the agents alongside the diff.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,68 @@
...
+<file>:<line>
+<what the code does, and the safe path>
+```
+
+Instruct every agent to:
+
+- report only what it can point at — a file, a line, and the text that breaks the rule;
+- skip code carrying `# rules-allow: <its-rule-id> — <reason>` on the line or its enclosing block, and
+  list what it skipped;
+- answer `none` plainly rather than reach for a marginal finding. A checker that always finds something
```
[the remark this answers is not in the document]
vertix: Fixed in 9291c1de. You are right that the two instructions contradicted each other — there was no response shape that could carry "nothing violated, but here is what I skipped".

Each agent now returns two sections, either of which may be empty:

```
FINDINGS
Rule <rule-id> violated:
...

WAIVERS
<file>:<line> — <the reason the waiver gives>
```

A rule whose every match is waived returns an empty `FINDINGS` and a populated `WAIVERS`, and Step 3 aggregates the two separately, so a waived rule can never be read as a rule that passed.

.claude/skills/add-rule/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+  separately whether to clean up now or let it bite on future changes — a rule that flags a lot of
+  existing code will make the next few reviews noisy, and the user should know that going in.
+- **Hits you disagree with.** The rule is too broad. Narrow it, or drop it.
+
+Report the sample honestly, including hits that make the rule look bad.
+
+## Step 6: Approve and append
+
+Show the final wording, the id, the hit count, and the sample. On explicit approval, append the rule to
+`CODE_RULES.md` in its section, then commit following the repo's conventions.
```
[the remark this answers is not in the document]
vertix: Fixed in 9291c1de. Real bug: the skill explicitly supports recalibrating an existing rule — `check-rules` sends false positives back here for exactly that — and Step 6 would then have appended a second heading with the same id.

Step 6 now branches: a new rule is appended, a revision of an existing id replaces that rule in place, so no id ever appears twice.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,68 @@
...
+what stops it from seeing a violation.
+
+## Step 1: Scope the diff
+
+```bash
+git diff --stat                                   # uncommitted work
+git diff $(git merge-base HEAD main)... --stat    # the whole branch
+```
+
+Use the uncommitted changes if there are any, the branch diff otherwise, or the files the user named.
```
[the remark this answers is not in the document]
vertix: Fixed in 9291c1de, by the same change that answers your staged-edits and untracked-files findings.

The scope is no longer a choice between working tree and branch. `git diff $(git merge-base HEAD main)` compares the merge base against the working tree, so committed, staged and unstaged changes are all in one diff, with untracked paths added from `git status`.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,68 @@
...
+```
+Rule <rule-id> violated:
+<file>:<line>
+<what the code does, and the safe path>
+```
+
+Instruct every agent to:
+
+- report only what it can point at — a file, a line, and the text that breaks the rule;
+- skip code carrying `# rules-allow: <its-rule-id> — <reason>` on the line or its enclosing block, and
```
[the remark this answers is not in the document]
vertix: Fixed in 9291c1de. Good catch — the waiver convention required a reason that says why the instance is correct, but nothing checked it, so `# rules-allow: hardcoded-keys — temporary` would have silenced a real finding while looking legitimate.

Agents are now told to report the violation anyway when the reason does not justify that instance, using your example verbatim as the illustration. Combined with the new `WAIVERS` section, an unjustified waiver is now louder than no waiver at all.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,80 @@
...
+The check runs in fresh subagents with no access to this conversation. That isolation is the point: the
+session that wrote the code knows why every line looks the way it does, and that knowledge is exactly
+what stops it from seeing a violation.
+
+## Step 1: Scope the diff
+
+```bash
+BASE=$(git merge-base HEAD main)
+git diff $BASE --stat                # committed, staged and unstaged, against main
+git status --porcelain | grep '^??'  # untracked files, which no diff shows
```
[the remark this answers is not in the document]
vertix: Fixed in f519d337. You are right, and right that the previous fix did not cover it: `--untracked-files=normal` is the default, so a wholly-new package showed up as one `?? dir/` entry and every file under it went unchecked — the exact case the earlier fix was supposed to catch.

Step 1 now uses `git status --porcelain --untracked-files=all`. Skipped `-z`: it guards against newlines in filenames, which is not a failure mode this repo has.

.claude/skills/add-rule/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+## Retiring a rule
+
+Rules only accumulate unless something removes them, and a stale rule dilutes attention across the
+whole file. Retire one when:
+
+- it has not fired in review for a long stretch, and a scan of the tree finds nothing it would catch;
+- it is waived more often than it is obeyed — count the `# rules-allow:` markers carrying its id:
+
+  ```bash
+  grep -rn 'rules-allow: <rule-id>' --include='*.py' .
```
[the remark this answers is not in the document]
vertix: Fixed in f519d337. The criterion counted a numerator with no denominator — five waivers looks damning until you notice five hundred compliant sites next to it.

The bullet now says to treat the grep as a prompt to look rather than as evidence, and to sample both the waived and the compliant uses before concluding: five against five hundred is a healthy rule with real exceptions; five against eight is a rule describing the exception as the norm.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,80 @@
...
+
+The check runs in fresh subagents with no access to this conversation. That isolation is the point: the
+session that wrote the code knows why every line looks the way it does, and that knowledge is exactly
+what stops it from seeing a violation.
+
+## Step 1: Scope the diff
+
+```bash
+BASE=$(git merge-base HEAD main)
+git diff $BASE --stat                                      # committed, staged and unstaged
```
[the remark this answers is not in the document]
vertix: Fixed in aabfc17a. The `MM` case is real: `git commit` records the index, so content can enter a commit having never appeared in the worktree the checker inspected.

Step 1 now takes both snapshots:

```bash
git diff $BASE --stat           # base -> worktree
git diff --cached $BASE --stat  # base -> index, what a plain commit records
```

with a note that they are identical unless a file was staged and then edited again, and that both are in scope when they differ.

AGENTS.md
```diff
@@ -0,0 +1,21 @@
...
+```
+Rule {name} violated:
+{details}
+```
+
+`{name}` is the rule's id — its heading in `CODE_RULES.md`. `{details}` names the offending code and
+the safe path out of it.
+
+Do not raise a violation when the offending line or its enclosing block carries a
+`# rules-allow: <rule-id> — <reason>` comment for that rule.
```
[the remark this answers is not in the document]
vertix: Fixed in 89f8fb9b. Correct, and a defect I introduced: the waiver-reason check went into `check-rules` in 9291c1de and I never carried it to the review path, leaving two enforcement paths that disagreed about what silences a finding — the worse of the two being the one that runs on every pull request.

`AGENTS.md` now requires the reason to say why the instance is correct before the violation is suppressed, and names the same `— temporary` example so both paths read alike.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,84 @@
...
+The check runs in fresh subagents with no access to this conversation. That isolation is the point: the
+session that wrote the code knows why every line looks the way it does, and that knowledge is exactly
+what stops it from seeing a violation.
+
+## Step 1: Scope the diff
+
+```bash
+BASE=$(git merge-base HEAD main)
+git diff $BASE --stat                                      # base -> worktree
+git diff --cached $BASE --stat                             # base -> index, what a plain commit records
```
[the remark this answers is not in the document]
vertix: Fixed in 1864b6d2. Correct — pushing sends HEAD, so a violation committed on the branch and corrected only in the index survives a check that never looks at HEAD.

Taking it as the close of this line rather than another patch. This is the fifth finding on this block, and each previous one added a case: staged, untracked, untracked directories, index-vs-worktree. Yours completes the enumeration instead of extending it — git keeps exactly three snapshots, and Step 1 now checks all of them plus untracked files:

```bash
git diff $BASE HEAD --stat      # committed — what a push sends
git diff $BASE --cached --stat  # index — what a plain commit records
git diff $BASE --stat           # worktree — what you see in the files
```

The prose says the set is closed and why, so the next reader sees a complete enumeration rather than four accumulated special cases. The accreted explanation those earlier fixes left behind is gone with it.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+
+The check runs in fresh subagents with no access to this conversation. That isolation is the point: the
+session that wrote the code knows why every line looks the way it does, and that knowledge is exactly
+what stops it from seeing a violation.
+
+## Step 1: Scope the diff
+
+```bash
+BASE=$(git merge-base HEAD main)
+git diff $BASE HEAD --stat                                 # committed — what a push sends
```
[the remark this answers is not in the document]
vertix: Fixed in 10f01381.

I said on the previous thread that a sixth finding on this block would be declined as diminishing returns. That was the wrong commitment and this finding is why: the gap you found did not exist before 1864b6d2 — my own fix added the HEAD comparison with `--stat`, which surfaces filenames and counts and discards the very content the comparison was added to expose. An agent reading clean files on disk would have no way to see a violation that lives only in HEAD.

The three commands now produce patches rather than diffstats, and the skill says why the agents need the text: when the snapshots disagree, the violating lines exist in one of them and nowhere on disk. `--stat` is noted as a separate thing to run when a file list is wanted for the report.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,90 @@
...
+hunt. A finding that does not name a rule is out of scope, however good it is.
+
+The check runs in fresh subagents with no access to this conversation. That isolation is the point: the
+session that wrote the code knows why every line looks the way it does, and that knowledge is exactly
+what stops it from seeing a violation.
+
+## Step 1: Scope the diff
+
+```bash
+BASE=$(git merge-base HEAD main)
```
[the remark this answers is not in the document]
vertix: Fixed in 14e99bbc — and it is worse than a missing-ref hypothetical. I checked this very worktree: local `main` is `88b27db4` while the true base is `736b34b5`, because local `main` has not been updated since someone merged upstream. So `git merge-base HEAD main` did not fail loudly; it silently returned an older base and would have widened every check to include commits already on the mainline.

Step 1 now resolves the base from a remote-tracking ref and stops rather than guessing:

```bash
BASE_REF=$(git rev-parse --verify --quiet upstream/main || git rev-parse --verify --quiet origin/main)
[ -n "$BASE_REF" ] || { echo 'no upstream/main or origin/main — name the base explicitly'; exit 1; }
BASE=$(git merge-base HEAD "$BASE_REF")
```

The prose notes that a remote ref can itself be behind, so fetch first when that matters.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,90 @@
...
+checks none. A narrow agent also fails visibly: reporting nothing is one rule cleanly checked, not
+nine rules half-checked.
+
+Each agent returns two sections, either of which may be empty:
+
+```
+FINDINGS
+Rule <rule-id> violated:
+<file>:<line>
+<what the code does, and the safe path>
```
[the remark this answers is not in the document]
vertix: Fixed in 14e99bbc. This is the usability hole the three-snapshot scope opened: a finding against HEAD sent the reader to a line that reads clean on disk, with nothing to say where the fix was owed.

Findings and waivers now both start with `<snapshot>` — `HEAD`, `index`, `worktree`, or `untracked` — and Step 3 preserves that field through aggregation rather than flattening to file and line.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,100 @@
...
+The check runs in fresh subagents with no access to this conversation. That isolation is the point: the
+session that wrote the code knows why every line looks the way it does, and that knowledge is exactly
+what stops it from seeing a violation.
+
+## Step 1: Scope the diff
+
+```bash
+BASE_REF=$(git rev-parse --verify --quiet upstream/main || git rev-parse --verify --quiet origin/main)
+[ -n "$BASE_REF" ] || { echo 'no upstream/main or origin/main — name the base explicitly'; exit 1; }
+BASE=$(git merge-base HEAD "$BASE_REF")
```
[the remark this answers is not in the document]
vertix: Fixed in 7b031d00, as a rewrite rather than a third guard.

You are right on the mechanism: an empty `BASE` does not error, it vanishes from the command line, and `git diff $BASE HEAD` silently becomes `git diff HEAD`. Every committed change disappears from the scope and the report comes back clean.

Two rounds ago this was one `git merge-base HEAD main`. The previous fix made it a ref check plus a merge-base call; yours would have made it a ref check plus a merge-base call plus an ancestry check. Three guards for one variable is the wrong shape, so the whole assignment is now one expression with one failure path:

```bash
BASE=$(git merge-base HEAD upstream/main 2>/dev/null || git merge-base HEAD origin/main 2>/dev/null)
[ -n "$BASE" ] || { echo 'no merge base — fetch, deepen a shallow clone, or name the base commit'; exit 1; }
```

Shorter than what it replaces and covers strictly more: missing ref, unfetched ancestry, and shallow clone all land on the same empty `BASE` and the same stop. The expansions are quoted as well, so the degradation you describe cannot happen even if the check were somehow bypassed.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+## Step 1: Scope the diff
+
+```bash
+BASE=$(git merge-base HEAD upstream/main 2>/dev/null || git merge-base HEAD origin/main 2>/dev/null)
+[ -n "$BASE" ] || { echo 'no merge base — fetch, deepen a shallow clone, or name the base commit'; exit 1; }
+
+git diff "$BASE" HEAD                                      # committed — what a push sends
+git diff "$BASE" --cached                                  # index — what a plain commit records
+git diff "$BASE"                                           # worktree — what you see in the files
+git status --porcelain --untracked-files=all | grep '^??'  # untracked — what no diff shows
```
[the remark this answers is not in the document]
vertix: Fixed in 2f8d9e40 — `|| true` added, since it costs one token and removes the ambiguity.

The grep behaviour is exactly as you describe. The impact is speculative here, though: this block is instructions an agent reads and runs, not a script executed by a wrapper that aborts on nonzero status. I applied it on cost, not on the failure scenario.

Marking the end of this line. Nine rounds have now landed on this one scope block: staged edits, untracked files, untracked directories, index vs worktree, HEAD, diffstat vs patch, base ref, merge-base availability, and now grep exit status. Every one was correct and I have declined none, but the block has stopped being where the risk is — the file's subject is code rules, and its scope resolution is now more thoroughly specified than the rules it exists to check.

Further findings on this block will get a reasoned decline unless they identify a rule check that is actually missed, rather than a shell edge case around a check that already runs. Findings on the rules themselves, the agent instructions, or the reporting shape are still very much wanted.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+## Step 1: Scope the diff
+
+```bash
+BASE=$(git merge-base HEAD upstream/main 2>/dev/null || git merge-base HEAD origin/main 2>/dev/null)
+[ -n "$BASE" ] || { echo 'no merge base — fetch, deepen a shallow clone, or name the base commit'; exit 1; }
+
+git diff "$BASE" HEAD                                      # committed — what a push sends
+git diff "$BASE" --cached                                  # index — what a plain commit records
+git diff "$BASE"                                           # worktree — what you see in the files
+git status --porcelain --untracked-files=all | grep '^??' || true  # untracked — what no diff shows
```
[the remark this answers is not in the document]
vertix: Declining this one, with the mechanism granted.

You are right that `git status ... | grep '^??' || true` swallows a `git status` failure as well as grep's no-match. The pipeline's status is grep's, and `|| true` covers both.

What I do not accept is the failure scenario. It requires `git status` to fail while the three `git diff` commands immediately above it succeed against the same repository. A repository whose metadata cannot be read fails all four, and the agent running them sees that plainly — there is no state where the checker sails past a broken repo on the strength of this one line.

That matters because of where the previous fix came from. The empty-search finding rested on a premise I flagged as speculative when I applied it: this block is instructions an agent reads and runs, not a script executed by a wrapper that aborts on nonzero status. I took it anyway because `|| true` cost one token. This follow-up asks me to spend two more lines hardening the fix to a speculative problem against a scenario that cannot occur — and the resulting code would be longer than what it replaces, for a case nobody has.

That is the shape `CODE_RULES.md` names `overspecific` in this very PR: generalise until it simplifies, and stop there. Adding a capture-and-validate step here goes past the stopping point.

Stated plainly one round ago and holding to it: this is the tenth finding on the same six-line block, and further shell edge cases around checks that already run will get the same answer. Findings that name a rule check actually being missed, or that touch the rules, the agent instructions, or the reporting shape, remain welcome.

Leaving this open for a human to close.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+name: check-rules
+description: Check the current changes against CODE_RULES.md and report violations. Use before pushing, or when asked whether a change breaks the repo's rules. Each rule is checked by an isolated agent with no session context.
+---
+
+# Check Rules
+
+Checks a diff against `CODE_RULES.md`. Nothing else — not a code review, not a style pass, not a bug
+hunt. A finding that does not name a rule is out of scope, however good it is.
+
+The check runs in fresh subagents with no access to this conversation. That isolation is the point: the
```
vyakunin: does clco actually allow truly-fresh subagents? (won't they inherit the parent's claude md and all the surrounding rules?)

vyakunin: (same applies to add-rule fresh subagent notes)

.claude/skills/add-rule/SKILL.md
```diff
@@ -0,0 +1,122 @@
...
+The rule will be applied by readers with no access to this conversation. Test that directly: hand a
+fresh subagent **only** the draft rule and two unlabelled code snippets — one violating, one clean —
+and ask which violates it and why.
+
+If it picks wrong, or its reasoning is not the reasoning you intended, the rule is unclear. Reword and
+repeat. Do not proceed on a rule that needed the conversation to be understood.
+
+## Step 5: Calibrate against the repo
+
+A rule that has never fired is a guess. Scan the existing tree — not just the recent diff — for code
```
vyakunin: this is not very clear. Do we expect the agent to read the whole codebase? It's prohibitively expensive at some point

.claude/skills/add-rule/SKILL.md
```diff
@@ -0,0 +1,122 @@
...
+the rule would flag, then **read a sample of the hits** (five, or all of them if fewer) and judge each
+one yourself.
+
+The count sizes the sample; the sample decides. Report both, and the fork they imply:
+
+- **Hits you agree with, few.** Calibrated. Land it.
+- **Hits you agree with, many.** The rule is right and the repo has a backlog. Land it, and raise
+  separately whether to clean up now or let it bite on future changes — a rule that flags a lot of
+  existing code will make the next few reviews noisy, and the user should know that going in.
```
vyakunin: I can't parse English grammar here

.claude/skills/add-rule/SKILL.md
```diff
@@ -0,0 +1,122 @@
...
+- **Hits you disagree with.** The rule is too broad. Narrow it, or drop it.
+
+Report the sample honestly, including hits that make the rule look bad.
```
vyakunin: I prefer to never have opus tell-tales words in core documentation, "honest(ly)" and "load-bearing" are some of them. In most of the cases they are not load-bearing.

.claude/skills/add-rule/SKILL.md
```diff
@@ -0,0 +1,122 @@
...
+Show the final wording, the id, the hit count, and the sample. On explicit approval, write it to
+`CODE_RULES.md` — a new rule is appended, a revision of an existing id replaces that rule in place, so
+no id ever appears twice. Then commit following the repo's conventions.
+
+## Retiring a rule
+
+Rules only accumulate unless something removes them, and a stale rule dilutes attention across the
+whole file. Retire one when:
+
+- it has not fired in review for a long stretch, and a scan of the tree finds nothing it would catch;
```
vyakunin: this is unenforcible fluff unless we provide machinery actually verifying it on certain cadence

.claude/skills/add-rule/SKILL.md
```diff
@@ -0,0 +1,122 @@
...
+- its waivers are out of proportion to its compliant uses. Count the markers, but treat the count as a
+  prompt to look, never as the evidence itself:
+
+  ```bash
+  grep -rn 'rules-allow: <rule-id>' --include='*.py' .
```
vyakunin: modern agents don't need guidance on how to grep stuff, it's dilutes cxt without any benefit

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+session that wrote the code knows why every line looks the way it does, and that knowledge is exactly
+what stops it from seeing a violation.
+
+## Step 1: Scope the diff
+
+```bash
+BASE=$(git merge-base HEAD upstream/main 2>/dev/null || git merge-base HEAD origin/main 2>/dev/null)
```
vyakunin: ditto, any agent already knows this in a basic setup

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+git diff "$BASE" --cached                                  # index — what a plain commit records
+git diff "$BASE"                                           # worktree — what you see in the files
+git status --porcelain --untracked-files=all | grep '^??' || true  # untracked — what no diff shows
+```
+
+Git keeps three snapshots — HEAD, the index, the worktree — and a violation can sit in any one of them
+while the others look clean: staged and then reverted, or committed and then fixed only in the index.
+Check all three, plus untracked files. That set is closed; there is no fourth place for code to hide.
+
+These commands produce patches, not summaries. The agents need the patch text: when the snapshots
```
vyakunin: ditto, any agent knows that

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+Narrow to the files the user named if they named any, and say which scope you used. A clean report over
+the wrong scope is worse than no report.
+
+## Step 2: One agent per rule
+
+```bash
+grep -n '^### ' CODE_RULES.md
+```
+
+Spawn one agent per rule, in parallel. Each gets exactly: the full text of **its own rule and no
```
vyakunin: if we get 20 rules this can get pretty expensive quickly, sure we want this?

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+FINDINGS
+Rule <rule-id> violated:
+<snapshot> <file>:<line>
+<what the code does, and the safe path>
+
+WAIVERS
+<snapshot> <file>:<line> — <the reason the waiver gives>
+```
+
+`<snapshot>` is `HEAD`, `index`, `worktree`, or `untracked` — which of the four the violating text lives
```
vyakunin: we focus so much on head/worktree/untracked. Is this intended? Can't the caller instruct the skill what's the actual change it wants to push (therefore needs to be reviewed)? I believe scope shouldn't be owned by the skill

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+in. Without it a finding against HEAD points at a line the reader opens and finds already clean, with
+nothing to say where the fix is owed.
+
+Instruct every agent to:
+
+- report only what it can point at — a file, a line, and the text that breaks the rule;
```
vyakunin: I don't understand what "report only what it can point at" is trying to convey here

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+- honour `# rules-allow: <its-rule-id> — <reason>` on the line or its enclosing block, and list every
+  waiver it honoured. A rule whose every match is waived returns an empty `FINDINGS` and a populated
+  `WAIVERS`, which is not the same as passing;
+- report the violation anyway when the waiver's reason does not say why that instance is correct.
```
vyakunin: where is this coming from? why singling out no-reson findings?

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+own rule and snapshot and the agents' format, so local output matches what Codex posts on the PR. State
+the scope and the rules checked, then list the collected `WAIVERS` separately — a rule silenced by a
+waiver must never look the same as a rule that passed.
+
+Report only. Fixing is a separate decision and the user makes it.
+
+## When a finding is wrong
+
+A false positive is information about the rule, not just about the code. If a rule produces findings
+you disagree with, take it to the `add-rule` skill and recalibrate or reword it there. Arguing with the
```
vyakunin: do we expect the skill to do it pro-actively or communicate back to the caller?

AGENTS.md
```diff
@@ -0,0 +1,23 @@
+# Positronic
+
+Python-native stack for real-life ML robotics: recording, datasets, policy training and on-robot inference.
```
vyakunin: nit: is inference deliberately the last in the list? How does this correspond to README?

AGENTS.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Code Review Rules
+
+Apart from your normal review, read `CODE_RULES.md` in the repository root and check the diff against
```
vyakunin, line 7: are we trying to make it codex- or cursor- friendly here? Why prose instead of standard framework-friendly directives?

AGENTS.md
```diff
@@ -0,0 +1,23 @@
+# Positronic
+
+Python-native stack for real-life ML robotics: recording, datasets, policy training and on-robot inference.
+
+## Code Review Rules
```
vyakunin, line 5: are these deliberately code _review_ rules, not code _authoring_ or _design_ rules?

vyakunin, line 5: I think generally this block is duplicative, just referring to check-rules skill should be sufficient since you have it

CODE_RULES.md
```diff
@@ -0,0 +1,59 @@
...
+Don't name anything — function, class, module, variable — after where it is used. Name it after what it
+does. Information about the callers must not leak into the name.
+
+A helper that recomposes a pose through a fixed transform is `change_frame`, not `to_policy_frame`: it
+has never heard of a policy.
+
+### hardcoded-keys
+
+Don't write the names of data fields as literals inside your code — dict keys, signal names, field
+paths. Take them as parameters instead, with a default for the usual name.
```
vyakunin: I am not sure that injection is always (or even often) the right remedy. Ofter literal is just a literal, it's not supposed to ever change. However, inlining the hardcode breaks DRY (don't repeat yourself) principle: this literal is the contract between different subsystems and hardcoding it everywhere make this contract implicit. The right way is usually to define a constant in a shared module accessible to all callers, so the contract becomes explicit and literal is used once. This, by the way, should be a lintable statically enforceable rule (we have cases of it already for a specific list of literals already having constants for them, should be possible to generalize to not introducing new string literals other than named consts.

CODE_RULES.md
```diff
@@ -0,0 +1,59 @@
...
+# Good — pose_key is a constructor parameter defaulting to 'ee_pose'
+return {**data, self._pose_key: change_frame(data[self._pose_key])}
+```
+
+### overspecific
+
+Don't bake assumptions about how a component will be used into its interface. Before adding a
+parameter, a branch, or a category to a signature, ask what else the component could legitimately be
+asked to do — many keys instead of one, several value types instead of one, the same operation in
+reverse. Where handling those uniformly makes the code **shorter**, the specific version is wrong.
```
vyakunin: "makes the code **shorter**"

i disagree with "shorter" criterion. Simpler/more readable is more important than line/char count. If you have a very generic and short signature, that the user/reader needs to check the implementation or usages to understand what it does – it's a bad outcome.

The goal of interface should be use convenience, "intuitivity". When all the uses naturally fall in place and you don't have to deliberate how to fit a particular case into the overly-generic framework not opinionated about the usages.

.claude/skills/check-rules/SKILL.md
vertix: Good question, and the answer is documented — subagents do start genuinely fresh:

> Each subagent starts with a fresh, isolated context window. It doesn't see your conversation history, the skills you've already invoked, or the files Claude has already read. Claude composes a delegation message that summarizes the task, and the subagent works from there. The exception is a fork, which inherits the parent conversation instead of starting fresh.

So no conversation history, no files this session read, no prior skill invocations. The one mode that does inherit is a fork, which this skill must never use.

The real hole is the sentence in the middle of that quote: the parent composes the prompt, and the parent is the one party that knows why the code looks the way it does. The platform cannot stop me writing "the pose keys are split because the two directions differ" into the delegation message, and at that point the isolation is gone — leaked by the author, not inherited from the session.

Reworded in eae68b23 to say that outright and to make the prompt a closed list rather than a list of things to avoid mentioning:

- the intro now states what the platform guarantees, names the fork exception, and says plainly that the prompt is the only channel in and you are the worst possible author for it;
- Step 2 specifies exactly two things that go into the prompt — the one rule verbatim, and the Step 1 patches — with everything else excluded by construction. The old wording paired a whitelist with a redundant "don't tell it X" blacklist; a closed list is checkable, a blacklist is not.
- `add-rule`'s cold-read test got the same treatment, since your follow-up applies there too.

On CLAUDE.md specifically: custom subagents do load it (only the built-in Explore and Plan skip it). Not a problem while its contents agree with `CODE_RULES.md`, and worth remembering if we ever want the agents genuinely rule-blind.

Leaving this open — it was a question, so it is yours to close.

vertix: Yes. Actually forking is rare, default option is fresh Subagent. We gonna strengthen the wording

.claude/skills/add-rule/SKILL.md
vertix: Fixed in 4a30f8f: calibration is no longer a gate on every rule. It is now a short section you run when a rule starts producing findings people argue with, so the cost is paid only when there is a reason to pay it.

vertix: Correction: the commit is `db1e9258`, not the sha named above.

.claude/skills/add-rule/SKILL.md
vertix: That sentence was inside the calibration fork removed in `db1e9258`, so it is gone rather than reworded.

.claude/skills/add-rule/SKILL.md
vertix: Agreed, and recorded rather than just fixed. Both instances you flagged went with the calibration cut in `db1e9258`; `3a80e037` adds a line to `CLAUDE.md`'s comments section banning empty intensifiers, with the test being that deleting the word leaves the sentence meaning the same thing. Swept the new files for the rest and dropped a "precisely".

.claude/skills/add-rule/SKILL.md
vertix: Removed in `c36e18ae` — the whole "Retiring a rule" section is gone, not just the criterion you flagged. The waiver-count and linter-supersedes criteria were answerable, but they were the only content left and a retirement procedure nobody triggers is the same fluff one level up. The frontmatter and intro no longer claim the skill retires rules either.

.claude/skills/add-rule/SKILL.md
vertix: Deleted in `e65fe797`. The two checks are now one sentence naming what to check and where; the commands are gone. The `sed` range was also a latent bug — it would have silently returned nothing the moment `pyproject.toml` gained a section between ruff and setuptools.

.claude/skills/check-rules/SKILL.md
vertix: Cut in `1752f3f4`, along with the `--stat` paragraph you flagged below. Step 1 went from thirty lines to eleven: the commands are gone and what remains is the requirement — collect HEAD, index, worktree and untracked contents, take the base from a remote-tracking ref rather than local `main`, stop when there is no merge base.

Worth noting where that block came from: Codex spent nine review rounds on it, each round adding a case, and it grew from one line to six plus four paragraphs. Every finding was correct in isolation. You and it were reviewing different things — it checked whether the script was right, you checked whether the script should exist.

.claude/skills/check-rules/SKILL.md
vertix: Gone in `1752f3f4` with the commands above it. That paragraph existed only because Codex caught me using `--stat` where a patch was needed — my error, not something a reader needs explained.

.claude/skills/check-rules/SKILL.md
vertix: Fair concern, leaving it as is for now and revisiting with data.

At three rules the cost is theoretical, and the fan-out is one sentence in the skill, so switching is cheap whenever we want it. The two alternatives are batching several rules per agent, or a pre-filter that only runs rules that could plausibly apply to the changed files. Which one is right depends on something we do not know yet — whether most rules never apply to most diffs. If that turns out to be true the pre-filter wins, because it cuts cost without giving up the one-rule focus.

Leaving open, since it is a question rather than a defect.

.claude/skills/check-rules/SKILL.md
vertix: Agreed and done in `888f901c`. Step 1 is now three short paragraphs: the caller names the scope, and with nothing named the skill checks the whole change as it would land if merged as it stands, working out what that means for the checkout in front of it.

The `<snapshot>` field went with it, from the finding format and from the aggregation step. It only existed to disambiguate a scope the skill had inferred; once the caller owns the scope there is nothing to disambiguate.

You found the root cause of something that had been eating the review. Codex spent nine rounds on that block — staged edits, untracked files, untracked directories, index versus worktree, HEAD, diffstat versus patch, base ref, merge-base availability, grep exit status. Every one of those was a consequence of the skill inferring scope for itself, so the whole class is gone rather than fixed.

.claude/skills/check-rules/SKILL.md
vertix: Reworded in `440d1271` to say it plainly: every violation carries a file and a line, best effort, because a finding with nowhere to point is not a finding.

.claude/skills/check-rules/SKILL.md
vertix: Agreed, removed in `114e4752` from both the skill and `AGENTS.md`.

It came from a Codex finding and I took it without asking the question you just asked. Any suppression can be abused — `noqa` and `type: ignore` included — and the answer to that has always been review, not a checker that grades excuses. `CODE_RULES.md` still says a waiver's reason should explain why the instance is correct; that is now a convention a human enforces, which is what it should have been.

.claude/skills/check-rules/SKILL.md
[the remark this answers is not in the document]
vertix: Reversing this on human review, in `114e4752`. Any suppression can be abused — `noqa` and `type: ignore` included — and the answer is code review, not a checker judging whether an excuse is good enough. The reason requirement stays in `CODE_RULES.md` as a convention with nothing enforcing it.

AGENTS.md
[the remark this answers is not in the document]
vertix: Reversing this on human review, in `114e4752`. Any suppression can be abused — `noqa` and `type: ignore` included — and the answer is code review, not a checker judging whether an excuse is good enough. The reason requirement stays in `CODE_RULES.md` as a convention with nothing enforcing it.

.claude/skills/check-rules/SKILL.md
vertix: Communicate back, never act. Fixed in `d9849bef`: the section now says to report the false positive and stop there, because only a human changes `CODE_RULES.md` — through `add-rule`, with explicit approval. The skill never edits a rule however wrong the rule looks from where it is standing.

The old wording ("take it to the add-rule skill and recalibrate it there") also contradicted Step 3, which already said report only.

AGENTS.md
vertix: Fixed in `b9c9c7f2`: the line is now the README's own, verbatim. The list was mine, matched nothing, and the ordering meant nothing.

AGENTS.md
vertix: `AGENTS.md` is the cross-tool convention Codex and Cursor both read, and prose under a `## Code Review Rules` heading is the format Codex documents — there is no directive syntax to use instead. The non-standard part is that our rules sit in a linked file rather than under that heading, which is what this PR is testing.

AGENTS.md
vertix: Right — they are authoring rules, and the heading undersells them. Fixed in `41cf5b46` by saying so outright: the rules govern writing code here, not only reviewing it, and the heading carries the review name because `## Code Review Rules` is the exact string Codex looks for. Renaming it would lose the reviewer.

vertix: Declining: Codex reviewing on GitHub cannot invoke a Claude Code skill, so `AGENTS.md` is the only channel that reaches it and `check-rules` only ever runs locally.

CODE_RULES.md
vertix: You are right that the rule named the wrong problem. Rewritten in `3928a2cb`: the damage is the implicit contract, not the inflexibility, and there are two remedies rather than one — a parameter when callers may legitimately differ, a named constant in a shared module when everyone must agree on the same name.

Not doing the lint for now. If it lands later the rule retires, since anything ruff can enforce does not belong in this file.

CODE_RULES.md
vertix: Agreed — "shorter" was reaching for something countable, and countable was the wrong goal. `def __init__(self, keys)` is shorter than the split version and tells the caller less.

Rewritten in `37ba623d`: the outcome to look for is simpler and easier for a human to read — fewer branches, often fewer lines, every use falling into place without deliberation. The stopping condition is now your failure mode stated directly: past the point where generalising simplifies, the signature stops having an opinion and the caller has to read the implementation to know what to pass. The rule also says outright that this is a matter of taste and does not reduce to a measurement.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+grep -n '^### ' CODE_RULES.md
+```
+
+Spawn one agent per rule, in parallel. Its prompt contains exactly two things:
+
+1. the full text of that one rule, verbatim from `CODE_RULES.md`;
+2. the Step 1 patches, plus the contents of any untracked files.
+
+Nothing else goes in: not the other rules, not what the change was for, not what you expect it to find,
+not why the code looks the way it does. The agent reads the repository itself for anything more.
```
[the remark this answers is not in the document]
vertix: Fixed in `931f57ee`. Real bug: "exactly two things" excluded the task and the response format, so an agent would have received a rule and a patch with no instruction to check anything and no output contract. The prompt now carries the task and the response format, plus exactly two pieces of *context* — the rule and the patch. The closed list applies to context, which is what it was always meant to constrain.

.claude/skills/add-rule/SKILL.md
```diff
@@ -0,0 +1,90 @@
...
+`<rule-id>` is kebab-case and names the violation, not the remedy — it appears in review comments as
+`Rule <rule-id> violated:` and in waivers as `# rules-allow: <rule-id> — <reason>`, so it must read as
+the thing that went wrong.
+
+## Step 4: Cold-read test
+
+The rule will be applied by readers with no access to this conversation. Test that directly with a
+subagent, which starts empty unless you fill it in. Its prompt contains exactly two things: the draft
+rule, and two unlabelled code snippets — one violating, one clean. Not which is which, not the incident
+behind the rule, not what you hope it will say. Ask which snippet violates the rule, and why.
```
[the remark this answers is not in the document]
vertix: Fixed in `931f57ee`, same defect as the `check-rules` prompt. The cold-read prompt now asks the question and carries exactly two pieces of context — the draft rule and the two unlabelled snippets.

CODE_RULES.md
```diff
@@ -0,0 +1,59 @@
+# Code Rules
+
+Judgment rules for this repository — the checks a linter cannot make.
+
+Every rule has an **id** — its heading. Cite it when reporting, fixing, or waiving a violation. Add,
+calibrate, and retire rules through the `add-rule` skill.
```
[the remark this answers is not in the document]
vertix: Fixed in `931f57ee`. Retirement was removed from `add-rule` on human review, and this line was a leftover reference to it; it now says add and recalibrate.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+`BASE` comes out empty and the run stops, rather than the expansion vanishing and each command
+degrading into a diff against HEAD that reports clean.
+
+Narrow to the files the user named if they named any, and say which scope you used. A clean report over
+the wrong scope is worse than no report.
+
+## Step 2: One agent per rule
+
+```bash
+grep -n '^### ' CODE_RULES.md
```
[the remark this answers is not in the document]
vertix: Obsolete. The skill no longer infers scope: on human review the caller now owns it, and with nothing named the skill checks the whole change as it would land if merged. Snapshot enumeration, base-ref resolution and rule retirement are all gone, so this finding no longer has a target.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+disagree, the violating lines exist in one of them and nowhere in the files on disk, so an agent that
+only reads the repository cannot see them. Add `--stat` separately if you want a file list for your own
+report.
+
+In the ordinary case the three agree and it is one patch. Pass the untracked files' contents alongside
+it — a brand-new file is where rules bite hardest and is the one thing no diff shows.
+
+The base is a merge base against a remote-tracking ref, never local `main` — that one goes stale the
+moment someone merges upstream and would quietly widen the scope to changes already on the mainline.
+One emptiness check covers every way this can fail: ref missing, ancestry not fetched, shallow clone.
```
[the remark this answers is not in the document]
vertix: Obsolete. The skill no longer infers scope: on human review the caller now owns it, and with nothing named the skill checks the whole change as it would land if merged. Snapshot enumeration, base-ref resolution and rule retirement are all gone, so this finding no longer has a target.

.claude/skills/add-rule/SKILL.md
```diff
@@ -0,0 +1,115 @@
...
+- Anything ruff or basedpyright can enforce. That belongs in `pyproject.toml`, where it runs
+  deterministically on every commit.
+- An aspiration — "keep interfaces clean", "name things well". Unfalsifiable, so it fires at random.
+- A symptom with no instruction attached.
+- Something so broad that most files violate it.
+
+## Step 1: Get the incident
+
+Ask what real code triggered this, and read it — the review comment, the commit, the function. A rule
+with no incident behind it is a preference; say so and stop.
```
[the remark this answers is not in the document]
vertix: Obsolete. The skill no longer infers scope: on human review the caller now owns it, and with nothing named the skill checks the whole change as it would land if merged. Snapshot enumeration, base-ref resolution and rule retirement are all gone, so this finding no longer has a target.

AGENTS.md
```diff
@@ -0,0 +1,25 @@
...
+
+Python-native stack for real-life ML robotics.
+
+## Code Review Rules
+
+Apart from your normal review, read `CODE_RULES.md` in the repository root and check the diff against
+every rule in it. The rules are not repeated here — that file is the whole set.
+
+They govern authoring as much as review: follow them when writing code here, not only when reviewing
+someone else's. The heading is named for the review case because that is the section name Codex looks
```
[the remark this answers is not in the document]
vertix: Fixed in `8fe7d3a9`. `CLAUDE.md` now opens with a line pointing at `CODE_RULES.md`, so the rules load in every authoring session rather than only when `check-rules` is invoked. That closes the gap between `AGENTS.md` claiming the rules govern authoring and nothing delivering them to the author.

.claude/skills/check-rules/SKILL.md
```diff
@@ -0,0 +1,83 @@
...
+
+```bash
+grep -n '^### ' CODE_RULES.md
+```
+
+Spawn one agent per rule, in parallel. Its prompt carries the task and the response format below, plus
+exactly two pieces of context:
+
+1. the full text of that one rule, verbatim from `CODE_RULES.md`;
+2. the change from Step 1, as patch text.
```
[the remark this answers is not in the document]
vertix: Not fixing this. The failure needs a caller to scope the check at a revision other than the checkout, and the remedy — reading every enclosing block from the scoped revision — is more machinery than the case earns. If it bites in practice it will be obvious from a waiver honoured against the wrong file.

Leaving open.

CLAUDE.md
```diff
@@ -65,6 +65,8 @@
 - Don't restate the code; if there is nothing to say beyond it, stay silent. Docstrings are plain statements of
   what the thing is
 - Comments wrap at 120 columns, same as code
+- No empty intensifiers, in comments or in docs — "honestly", "load-bearing", "genuinely", "truly", "precisely".
+  Delete the word; if the sentence still means the same thing, it was never doing any work
```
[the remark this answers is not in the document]
vertix: Fixed in `8fe7d3a9`: `CLAUDE.md`'s "truly unavoidable" and `docs/architecture.md`'s "load-bearing consequences" are reworded, along with three uses of "genuinely" in `CLAUDE.md` that you did not cite but that failed the rule's own test — delete the word, and the sentence means the same. The rule and the repo's own prose now agree.

CLAUDE.md
```diff
@@ -58,20 +60,28 @@
...
-- Don't restate the code; if there is nothing to say beyond it, stay silent. Docstrings are plain statements of
-  what the thing is
+- A docstring states what the thing is, and nothing else. Behaviour worth pinning down belongs in a test name,
+  a design decision in a TODO or a commit message, a usage example once in the module docstring
+- Leave a comment only for what the code cannot say: an invariant, a constraint from outside the file, a reason
+  the obvious alternative fails. Dry and factual. If it restates the code, delete it
+- Be economical: a class docstring runs a few lines, a method's one, and most methods need none. Past that it is
+  teaching rather than stating — cut it, or move the content to the home named above
 - Comments wrap at 120 columns, same as code
+- No empty intensifiers, in comments or in docs — "honestly", "load-bearing", "genuinely", "truly", "precisely".
```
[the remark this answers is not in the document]
vertix: Fixed in `8418f89e` — and you were right that the four you named were not the whole set.

Swept every file in the repo for the five banned words, not only the documentation. Twelve sites, all of them clean deletions where the sentence means the same thing without the word:

- `.claude/skills/address-review/SKILL.md` — lines 18, 38, 103, 225 (the four you found)
- `.claude/skills/address-review/watch.sh:68` — "genuinely no checks yet"
- `.claude/skills/polish/SKILL.md` — lines 18, 145, 403
- `positronic/offboard/client.py:18` — "a truly stalled/half-open connection"
- `positronic/simulator/libero/validate.py:153` — "the load-bearing, non-circular measurement"
- `utilities/check_basedpyright_baseline_ratchet.py:88` and its test at line 93 — "a genuinely new file"

The only remaining hit is `CLAUDE.md:76`, the rule naming the words.

PR #562, opened 2026-08-04
Add `positronic-server-probe` to render a viewer page and report what it showed

v-positronic:
The dataset viewer fails visually — a video panel stuck on its loading spinner, a plot that never fills — and none of it reaches the server log. So a report like "the video is black" cannot be checked by reading code or curling the endpoint; the page has to be rendered and looked at.

`utilities/viewer_probe.py` drives the real page in a headless Chrome, waits for the recording to stream in, steps the time cursor, saves a screenshot per step and prints the browser console.

```
uv run --script utilities/viewer_probe.py --url https://host:8400/episode/5
uv run --script utilities/viewer_probe.py --rrd ./episode.rrd --viewer https://host:8400
```

`--rrd` serves a recording from disk to the viewer's bundled rerun. That is the mode that settles causes: rebuild a recording with parts left out until the symptom flips. It is how the current "molmo videos never render" report was traced to an 866-element `sim_state` logged as 866 scalar series, after B-frame H.264, the container frame rate and a serve-time re-encode had each been suspected and cleared.

A debugging tool's dependencies do not belong in the library's, so this is a PEP 723 script rather than a package module: `playwright` reaches neither `pyproject.toml` nor the lockfile, and nothing in `positronic` can import it. The browser is the system Chrome — the bundled Chromium ships without H.264, so video panels stay blank there whatever the recording holds.

Refs Positronic-Robotics/internal#128.


vertix:
@codex review

vertix:
@codex review

v-positronic:
Relocating to the internal platform repo (see Positronic-Robotics/platform#54): this tooling is not for the public positronic side.

Issue #534, opened 2026-07-28
The server should own inference recording: declare the taps, report into them


vertix:
Inference recording is authored on the wrong side. The rig builds taps around a stack it did not
author, and the server records separately into a file nobody fetches. Since #497 the server owns the
policy pipeline and declares the local half; it should own the recording of it too.

## Today

Two independent recordings:

- client — `positronic/policy/remote.py:221`: `rec.tap('raw') | stack | rec.tap('server')`
- server — `positronic/offboard/server.py:260`: taps `raw` / `inference`, into the server's own
  `recording_dir`

The encoded observation the model consumed and the model's raw output land in a different file, on a
different machine, on a different clock — and in practice nobody fetches the server-side one.

Three consequences:

1. **`raw` duplicates the dataset.** The ds writer already records the rig's cameras and state at
   control rate, video-encoded. The `raw` tap writes the same signals again as loose full-resolution
   JPEGs — ~98% of the bytes in the file.
2. **`raw` fires on every control tick.** It sits above `ChunkedSchedule`, which returns `None` on
   ticks that run no inference, so it logs ~13 observations per inference instead of one. Two runs of
   the same rig, same episode length, same cameras:

   | entity | 24 Jul (pre-#497) | 27 Jul (post-#497) |
   |---|---|---|
   | `/raw/image.wrist`, `/raw/image.exterior` | 52 each | **687 each** |
   | `/server/image.*` | — (single tap) | 8 each |
   | file size | ~4 MB | **227 MB** |

   `s3://runway/inference/270726/rrd/` is 677 MB for two real episodes, against ~4 MB/episode before.
3. **The server's half is unreachable.** It is on the server's disk, on the server's clock, in its own
   file.

## End state

The `local_stack` spec gains tap points. The server says where the seams are — in practice one, at
the wire — and the rig builds the recording from that declaration instead of bracketing a stack it did
not author. The server ships its own entries into that stream, stamped against the client's timelines,
so one `.rrd` shows the whole round trip.

What follows:

- The client stops authoring taps. `recording_dir` stays a rig-side setting — where the file goes —
  while what goes into it is the server's call.
- `raw` disappears; the dataset already covers it.
- The volume problem stops existing rather than being suppressed: a tap at the wire seam sits below
  the scheduler, so a skipped tick has nothing to log. No gate, no frame-rate cap.
- The server stops writing its own file.

## Open questions

[...]

PR #564, opened 2026-08-04
Log recorder frames only on ticks that ran inference

vertix:
## Summary

Fixes part 1 of #534: the `raw` tap sits above `ChunkedSchedule`, so it saw every control
tick (~14 Hz) instead of one observation per inference, and inference `.rrd` files grew ~50x
(4 MB -> 227 MB per episode; full-resolution JPEG frames were ~98% of the bytes).

- `_RecordingTapSession.__call__` logs *after* the inner call returns, so a tap can tell an
  inference from a tick the scheduler skipped (`ChunkedSchedule` returns `None` for those).
- Frames are gated on that; numeric entries still log every tick, so the state trace
  (`robot_state.q`, `ee_pose`, `grip`) keeps full control rate at kilobyte cost.
- The blueprint now goes out on the first tick that ran inference rather than at step 0 —
  that is the first tick every inner tap logged through, and the only one with image paths
  registered.
- Nothing moves in the pipeline, so file content is otherwise unchanged: inner taps just log
  before the outer one within a tick, and everything is timestamped.

Part 2 of the issue (server-side entries reported into the client's tap) is not touched.

No frame-rate cap is added: volume is tied to the inference rate, which is what "restore the
default" means, but it does not bound anything — a server inferring at 5 Hz brings the growth
back.

## Test plan

- `positronic/policy/tests/test_recording.py::test_frames_follow_inference_not_control_ticks`
  ticks a `rec.tap('raw') | ChunkedSchedule()` pipeline twice inside one chunk and asserts one
  frame, two numeric entries. Verified it fails without the gate.
- Full suite: 945 passed, 7 skipped. Not verified on a real rig.

### DIFF
--- a/positronic/policy/recording.py
+++ b/positronic/policy/recording.py
@@ -19,6 +19,12 @@ correlated recording::
 - the ``server`` tap (innermost, next to the remote policy) logs the observation as
   sent to the server and the chunk as received back.
 
+Both are logged once the inner call returns, so a tap above a scheduling wrapper can tell
+an inference from a control tick the wrapper skipped (``ChunkedSchedule`` returns ``None``
+for those). Frames are written only on ticks that ran inference, which ties file size to the
+inference rate; numeric entries are written on every tick, so the state trace keeps the full
+control rate.
+
 Each entity is stamped on the timelines given by ``timelines`` (a mapping of rerun
 timeline name to observation key; by default ``wall_time`` and ``obs_time``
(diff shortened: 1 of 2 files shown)


vertix:
Closing: gating the client-side tap treats the symptom. The rig should not be authoring taps around a stack it did not author at all — the server owns the pipeline and should own its recording. Reworked #534 around that.

Issue #514, opened 2026-07-25
Drop the deprecated `--policy.local` and `--policy.compress_images` overrides


vertix:
`RemotePolicy` carries two deprecated operator overrides, kept only so a rig on a current build can still drive an inference server deployed before the server-declares-the-wire change:

- `--policy.local` — stands in for the `local_stack` a server would declare
- `--policy.compress_images` — stands in for `remote(compress_images=...)`

Both are resolved by `_operator_override` in `positronic/policy/remote.py`, which carries the TODO. The rule today: an override applies only when the server declares nothing of its own, logs a deprecation warning when it does apply, and raises against a server that declares.

**What removes them:** every deployed server running a build that declares its own wire — including the customer-facing Modal deployments, which live outside this repo. Once that holds, delete the two parameters, `_operator_override`, and the `--policy.local` / `--policy.compress_images` lines from `docs/inference.md`.

Note that a server whose pipeline uses the bare `remote` marker declares `compress_images: false`, which is a declaration — so a rig passing `--policy.compress_images=True` at it will raise. The fix in that case is to change the served pipeline to `remote(compress_images=True)`, not to keep the override alive.

Introduced in #497.

PR #567, opened 2026-08-05
Let the served pipeline declare where the rig records

vertix:
Step 1 of #534: the client half. The server shipping its own entries into the rig's stream is a
follow-up, tracked by a `TODO(#534)` on the code it replaces.

## Summary

`RemotePolicy` bracketed the server-declared stack with taps it authored itself —
`rec.tap('raw') | stack | rec.tap('server')`. The outer one sat *above* `ChunkedSchedule`, which
returns `None` on ticks that run no inference, so it logged every control tick: ~13 observations per
inference. Two runs of the same rig, same episode length, same cameras: 52 frames per camera before
#497, **687** after; ~4 MB per episode became **227 MB**.

- A `Tap(name)` names a seam in the pipeline itself — `... | Tap('wire') | remote | codec | source` —
  and travels to the rig inside the `local_stack` declaration, joining `WIRE_WRAPPERS` like any other
  entry.
- `from_spec(node, tap=...)` builds each declared seam through a factory, so where a recording lands
  is the builder's call rather than the declaration's: the rig points it at its `Recorder`, and a rig
  that records nowhere passes straight through. The tap is the only entry whose behaviour is not
  fully described by its args, so it is the only one routed this way; the vocabulary table stays
  closed.
- The client stops authoring taps. `recording_dir` still says *where* the file goes; *what* goes into
  it is the served pipeline's call. `raw` is gone — the dataset writer already records the rig's
  cameras and state at control rate.
- The volume problem stops existing rather than being suppressed: a seam below the scheduler only
  sees the ticks that ran an inference. No gate, no frame-rate cap.
- All six vendor pipelines name their wire seam. A rig given `recording_dir` against a server that
  names none records nothing and says so at connect.
- The legacy stand-in for a server that declares no stack at all (`TODO(#514)`) keeps a wire seam, so
  `--policy.recording_dir` does not silently stop working against those servers.

Left alone deliberately: the server's own `recording_dir` and its hardcoded `raw`/`inference` taps.
Those go once a `Tap` in the remote half can report into the rig's stream, which needs the wire
protocol change this PR does not make.

## Test plan

- `uv run --locked --extra dev pytest --no-cov` — 950 passed.
- New tests: the declared seam records under the name it was given; a declared seam logs once per
  inference across three gated ticks (the regression); `recording_dir` against a declaration with no
  tap warns and writes nothing; the legacy path still records; an unbound `Tap` leaves the policy
  untouched; `from_spec` routes the seam through the caller's factory; `tap` joins the wire-name
  table check.
- End to end over a real websocket (`PolicyServer` declaring `chunked_schedule | tap`, a real
  `RemotePolicy` recording), 20 control ticks / 4 inferences:

  | recording shape | image logs | `.rrd` bytes |
  |---|---|---|
  | declared tap, at the wire | 4 (`wire/image.exterior`) | 46,560 |
  | tap wrapped around the stack (this PR removes) | 20 (`raw/image.exterior`) | 110,152 |

- Not verified: a full eval against a real vendor server with a loaded checkpoint.

### DIFF
--- a/docs/connect-your-model.md
+++ b/docs/connect-your-model.md
@@ -141,12 +141,11 @@ Which command type your model produces is decided by its codec.
 
 ## Debugging with recordings
 
-When a run doesn't produce the result you expected, it helps to record exactly what crossed the boundaries between the robot, the codec, and the model. Recording is itself a policy wrapper — `Recorder` in [`positronic/policy/recording.py`](../positronic/policy/recording.py) — that taps into any client pipeline; the built-in servers expose it via `--recording_dir`. It writes one [rerun](https://rerun.io) file per episode with two layers:
+When a run doesn't produce the result you expected, it helps to record exactly what crossed the boundaries between the robot, the codec, and the model. Recording writes one [rerun](https://rerun.io) file per episode; where the recording is taken is a property of the pipeline, not of whoever runs it.
 
-- **`raw`** — the observation and action as they appear on the wire.
-- **`inference`** — the same episode *after* the codec: the encoded observation the model received and the raw actions it produced.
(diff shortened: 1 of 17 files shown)


vertix:
Closing unmerged. Step 1 of #534 is still unsolved in main — no `Tap` exists and the recording-volume blowup stands. Branch kept at `vertix/positronic-open:issue-534-declared-taps` if this gets picked up again.

PR #600, opened 2026-08-10
Skip a command pose present only as a null rename alias

v-positronic:
`ChangeEEFrame`'s episode transform converts only the pose keys the episode carries a signal for, rather than every key present in it. A key present with a `None` value passes through untouched.

## Why

`cfg.ds.internal._RENAME_ROBOT_COMMAND` backfills the canonical `robot_command.pose` from the legacy plural name with `Get('robot_commands.pose', None)`, so a dataset carrying neither name still holds the key — with `None` under it. Membership was the whole guard, so the transform wrapped that `None` in an `Elementwise` and handed back a signal with nothing inside. The failure then surfaced at the first dereference, as `TypeError: object of type 'NoneType' has no len()`, with nothing in the traceback naming either this transform or the rename.

Two other shapes are arguable:

- **Raise here rather than skip.** Absence already has a meaning in this transform: a key missing outright is skipped, which `test_training_encoder_skips_absent_command_pose` pins. The backfill's `None` is that same absence in a different spelling, so raising would make one spelling of "no command pose" fatal and the other benign.
- **Fix the backfill so absence stays absent** — drop the key instead of setting it to `None`. That is the narrower defect, with a wider blast radius: the alias is read by every consumer of `REAL_ROBOT_TRANSFORM` and `SIM_ROBOT_TRANSFORM`, so changing what it emits is a change to the dataset-transform contract. The guard here holds whoever produced the `None`.

What it does not cover: the null alias still reaches every other consumer of those transforms. This makes the frame codec tolerant of it, nothing more.

## Verification

The new test fails on the unguarded code with the `TypeError` above and passes with the guard. Full suite `uv run --locked pytest --no-cov` — 1038 passed, 8 skipped. `basedpyright` 0 errors, baseline ratchet clean, `ruff check` and `ruff format --check` clean.

## Refs

`Positronic-Robotics/internal#330`

### DIFF
--- a/positronic/policy/codec.py
+++ b/positronic/policy/codec.py
@@ -543,11 +543,9 @@ class ChangeEEFrame(Codec):
                     'episode a codec already moved would train on a frame the checkpoint does not declare'
                 )
             derived: dict[str, Any] = {obs_keys.EE_FRAME: FromValue(codec._transform.as_vector(_QUAT))}
-            # Membership alone is not enough: a joint-only dataset carries no legacy pose to rename, and the
-            # backfill (``cfg.ds.internal._RENAME_ROBOT_COMMAND``) surfaces that absence as a present key holding
-            # ``None``. Wrapping it would defer the failure to whoever dereferences the signal.
-            present = (key for key in codec._keys if key in episode and episode[key] is not None)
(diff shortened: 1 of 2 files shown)


vertix:
@codex review

PR #607, opened 2026-08-11
Report the MolmoSpaces end-effector pose in the robot frame, not the scene world frame

v-positronic:
## What

The MolmoSpaces env's whole Cartesian vocabulary — the reported eef pose, the IK target, the delta composition — moves from the ProcTHOR scene world frame to the robot base frame. MuJoCo still reads and writes the grasp site in world, so `_fk`/`_ik` convert through the base pose at that boundary and nowhere else. The parity reference and the names that said "world" follow.

## Why

A DROID-trained checkpoint reads `robot_state.ee_pose` as a robot-base-frame pose — on the real rig `|xyz| ~= 0.68 m`. This env was reporting the grasp site in the house frame, `|xyz| ~= 9.24 m`, which is outside anything such a checkpoint has ever seen as proprioception, so the eef channel was noise to it.

Robot frame rather than any other convention because it is what MolmoSpaces itself reports a TCP in: its own sensors and kinematics read `leaf_frame_to_robot` and compose `base.pose @` it only where world is wanted explicitly, and its LeRobot conversion feeds `obs/extra/tcp_pose` as the eef. This mirrors native semantics rather than inventing one.

The alternative is to convert on the positronic side — in the adapter or the codec — and leave the env world-framed. Rejected: a Cartesian command arrives in the frame the observation reports, so a client-side conversion splits that pair across the wire and leaves the server solving IK against a frame nobody reports. One frame, converted where MuJoCo is touched, is the shape that keeps command and observation together.

**Shortcoming, and why this is not a merge as it stands:** it changes what `robot_state.ee_pose` means in recorded episodes. What recorded episodes should carry is a decision of its own and this PR does not take it.

## Verification

- `validate.py` against a live benchmark scene: FK identity, IK round-trip over 16 sampled reachable targets, Cartesian hold, and command-contract coverage all pass. Its first run is what caught the substantive defect here — `_fk`/`_ik` still spoke world while the observation had moved, so FK identity was off by the base offset.
- The native-vs-positronic parity check passes on the pinned episode: 456 frames, terminating at MolmoSpaces' own horizon of 455.
- Both recorded pi05 rollouts replay bit-identically (max sim-state deviation within the 1e-6 budget).
- 41 unit tests pass; `ruff check`/`format` clean.

## Refs

No ticket; a probe for the Runway gyros rollout, on top of the MolmoSpaces integration branch.


vertix:
@codex review

PR #613, opened 2026-08-11
Reduce a run whose `eval.pass` span never closed over its episodes

v-positronic:
## What

`positronic eval timing-report` now reduces a run that was killed or preempted mid-pass. Where no `eval.pass` span closed, the wall window falls back to the span of the run's complete episodes, grouped by the pass span they name as their parent, so two killed runs appended to one directory keep a window each and the dead wall between them counts for neither.

Every share, and the real-time factor, is a fraction of one window, and a new `window` field names which — `W_pass` or `W_episodes`. `PassReport.wall_pass_s` becomes `wall_s` for the same reason: the field no longer always measures a pass.

## Why

`timing_report` gated on a closed `eval.pass` span, which is only written when a pass finishes. A killed run therefore reduced to nothing however many complete episode spans were on disk — and the orphan filter dropped those episodes a second time, since every one of them names a parent span that was never written. A preempted box is exactly the case whose timings someone wants to read.

The alternative shape is to reduce the same numbers and simply omit the pass-derived ones — drop `W_pass` and the wall split, keep the inference latencies. Rejected: that is most of the value gone for the case this fixes. So the figures are recomputed against the fallback window instead, and the window rides with them rather than being implied. `W_episodes` is narrower by construction — it excludes whatever ran either side of the episodes — so a share of it is not a share of a pass, and nothing here silently relabels one as the other. The report says which window it used in its own rendered body as well as on the console, because the console warning is lost the moment the text is pasted anywhere.

`no telemetry under <dir> (recorded without --timing?)` now fires only when there is genuinely no telemetry. Spans carrying neither a closed pass nor an episode name what is missing, rather than sending a reader after a flag they did set.

## Verification

`pytest positronic/cli/eval/tests/test_timing_report.py` — 27 pass, including four new cases: the fallback window and its shares, two killed runs in one directory getting a window each, the refusal that names what is missing, and the rendered body naming `W_episodes` with no `W_pass` anywhere in it. `ruff check`/`format` clean.

## Refs

Closes internal#353.

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -105,9 +105,9 @@ positronic eval timing-report <run_dir> [--gpu_policy_log <nvidia-smi dmon log>]
 Reduces the sidecars under `<run_dir>/telemetry/` into a pass report and writes `timing_summary.json` beside
 the input (`<run_dir>` may be an `s3://` URI). It reports:
 
-- the **wall split** — each phase's share of `W_pass` (reset / env.step / policy wait / record IO / overhead /
-  between-episodes), summing to 1; the policy-wait share also carries the sizing figure derived from it, how
-  many such evals one policy server could keep fed;
+- the **wall split** — each phase's share of the report's wall window (reset / env.step / policy wait / record
+  IO / overhead / between-episodes), summing to 1; the policy-wait share also carries the sizing figure derived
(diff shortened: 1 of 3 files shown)


vertix:
@codex review

PR #614, opened 2026-08-11
Give every `production` endpoint its own credentials

vertix:
## Summary

`production` mapped a name to a URL and built every `RemotePolicy` without `headers`, so pointing one of its endpoints at a gated server got a 401 on the model list and a refused WebSocket handshake. Follow-up to the thread on #565, which deferred the fix because it needs a decision: one bearer header shared by the whole dict, or credentials per endpoint.

**Per endpoint.** Sharing is wrong here beyond being the wrong shape — the endpoints a run routes between need not share a token, and `phail_multiple` names four plaintext `ws://` boxes that would each receive a secret they have no use for.

So an endpoint is a whole `remote` config rather than a URL, and everything one server needs is set on that server alone:

```bash
uv run positronic-inference phail \
  --policy.endpoints.groot=.nebius_remote \
  --policy.endpoints.groot.url=https://<managed-url>
```

`groot` gets `Authorization: Bearer <token>`; the other four stay open. `.authed_remote` works the same way against an exported `AUTH_TOKEN`.

- `production` no longer builds connections — it routes. `endpoints` is `dict[str, RemotePolicy]`.
- `recording_dir` leaves `production` with that: it is a `RemotePolicy` argument, reached as `--policy.endpoints.<name>.recording_dir`. There is no longer one run-wide flag for a whole fleet — configuronic instantiates dict values unconditionally, so keeping it would have meant the unstated contract "every endpoint config accepts `recording_dir`". `--policy.recording_dir` on the single-endpoint `.remote` path is unaffected.
- Adding an endpoint now names its config: `--policy.endpoints.groot=@positronic.cfg.policy.remote` and then its arguments. Repointing an existing one is unchanged (`--policy.endpoints.groot.url=...`).

## Test plan

- `positronic/cfg/tests/test_policy.py` — new, covering `production`, which had no tests. The first case swaps one `phail_multiple` entry for `.authed_remote` and asserts that endpoint alone carries the bearer header while the other four stay open; it exercises the relative-import form the CLI takes.
- `uv run --locked pytest --no-cov` — 1116 passed, 9 skipped.
- `uv run positronic-inference phail --help` renders each endpoint as its config.

### DIFF
--- a/positronic/cfg/policy.py
+++ b/positronic/cfg/policy.py
@@ -101,11 +101,11 @@ def production(
     """Routes each episode to one of several remote endpoints, each named for CLI overrides.
 
     An endpoint is a whole `remote` config rather than a URL, so what one server needs stays that
-    server's and reaches no other: `--policy.endpoints.groot.url=ws://desktop:8000` repoints one and
-    `--policy.endpoints.groot.headers='{"Authorization": "Bearer <token>"}'` credentials it. Adding one
-    names the config it is built from, `--policy.endpoints.groot=@positronic.cfg.policy.remote`, and
-    then its arguments. `weights` name the same endpoints and set their sampling odds; endpoints left
-    out of it weigh 1.0.
+    server's and reaches no other — `--policy.endpoints.groot.url=ws://desktop:8000` repoints one, and
+    `--policy.endpoints.groot=.nebius_remote` with its own `.url` moves that one alone behind a token
+    while the rest stay open. A name not already here is built from the config it is given,
+    `@positronic.cfg.policy.remote`. `weights` name the same endpoints and set their sampling odds;
+    endpoints left out of it weigh 1.0.
     """
     if not endpoints:
         raise ValueError(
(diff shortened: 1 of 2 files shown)


vertix:
Closing: the gap this fixes is latent and not worth the churn to reach for.

Every endpoint `production` and `phail_multiple` name is a self-hosted box that serves open, and the Nebius workflow is single-endpoint by construction — serve one checkpoint, eval it — which `.authed_remote` and `.nebius_remote` already cover. Reaching the 401 needs a gated endpoint inside a multi-endpoint config, which nothing is set up to do.

The structural point stands and is the reason to revisit: while `endpoints` is `dict[str, str]`, giving one endpoint a header, a timeout, or a recording directory means adding a second dict to `production` keyed by the same names. Whenever a served endpoint does join the fleet, the branch `claude/gallant-volhard-ceb5ef` carries the change — `endpoints` becomes a name-to-`remote`-config map, so a credential set on one endpoint reaches no other.

One factual correction from the work, which holds either way: `production` accepted a `recording_dir` that nothing has ever set. The fleet's recordings come from the servers, which bind their own `recording_dir` in each vendor's deployment preset.

PR #616, opened 2026-08-12
Expand `~` in the filesystem paths a CLI entry point opens

### LINKED ISSUE #615: `eval timing-report --run_dir` does not expand `~`, and blames the recording for it
`positronic eval timing-report --run_dir=~/evals/timing` fails as if the run carried no telemetry:

```
ValueError: no telemetry under ~/evals/timing/telemetry (recorded without --timing?)
```

The run *was* recorded with `--timing` and the sidecars are there. `~` simply reaches `Path` unexpanded, so
`root / TELEMETRY_SUBDIR` names a directory that does not exist. Passing the absolute path works:

```bash
positronic eval timing-report --run_dir=/home/vertix/evals/timing
```

The error message names the one cause that is not the actual one, so the reader goes looking at how the run was
recorded rather than at the path they typed.

`timing_report` in `positronic/cli/eval/timing_report.py` builds the local path with
`Path(run_dir)`, with no `expanduser`. The `s3://` branch beside it goes through `pos3.download`, and
`eval run --output_dir` goes through `pos3.sync`, which is why the same `~` works there — a
`~`-relative path is the documented form in `docs/evaluation.md`, where the quickstart writes to
`~/evals/libero` and would be read back by this command.

Worth checking whether any other CLI entry point takes a filesystem path straight to `Path` without the same
treatment, rather than fixing this one call site alone.

Found while running the real-policy eval pass for #591; unrelated to that change.

v-positronic:
`positronic eval timing-report --run_dir=~/evals/timing` failed as though the run carried no telemetry.
The `~` reached `Path` unexpanded, so the reduce looked under a literal `~/evals/timing/telemetry` — and
the error it raised named the one cause that could not be it, sending the reader to check how the run was
recorded rather than the path they typed. A `~`-relative directory is what `docs/evaluation.md` tells a
reader to write, and `eval run --output_dir` accepts one because `pos3.sync` expands it.

`timing_report` now expands both of its local paths, and its failure splits by cause: no telemetry
directory at all (not a run directory, or recorded without `--timing`), against a directory carrying no
spans (a run killed before its sidecars flushed). The rest of the class comes with it — the plot output,
the arm generator's URDF and target, the kinematics debug recording, the server's RRD cache (whose default
expanded itself while a caller's `~` did not), the GR00T venv, and the MuJoCo model and Kinova URDF.

The expansion goes where the path is opened, which is what `pos3` and `LocalDataset` already do and why a
`~` output directory works for `eval run` today. For the MuJoCo model and the Kinova URDF that point is
the class that loads them, so one line covers every CLI feeding them rather than each CLI repeating it.
The `s3://` branch is left alone deliberately: the URI is the remote store's to parse. A shared
`local_path()` helper was the alternative and buys nothing — `Path(x).expanduser()` is the idiom already
in `local_dataset` and the vendors. The durable closure is `configuronic` coercing a `Path`-annotated
parameter, which would also let these signatures say `Path` instead of `str`; that is a change in that
repo, not this one.

The sweep read every `cfn.cli` entry point's path-shaped parameters, from an AST walk over the repo,
through to whatever opens them. Two groups are deliberately untouched. Anything reaching `pos3.sync` /
`download` / `upload` or `LocalDataset*` is expanded there already: `eval run --output_dir`, every vendor
train and convert output, `probe`, `inference stats`, `fake_dataset_generator`, `convert_ds`,
`validate_server`, `migrate_remote`, `release_phail`. And a model id a third-party loader resolves
(`lerobot train --base_model`, `lerobot_0_3_3`'s `act --checkpoint_path`) is an id that may also be a
path, not a path this repo opens.

Verified with `uv run --locked pytest --no-cov` (1107 passed) and `pre-commit run` over the changed files.
Four tests in `positronic/cli/eval/tests/test_timing_report.py`, each confirmed to fail before the
change: a `~` run dir and dmon log resolve against `HOME`; an `s3://` run dir reaches `pos3` verbatim,
which is the boundary an over-broad expansion would break; and each of the two failures names its own
cause.

Fixes #615.

### DIFF
--- a/positronic/cli/eval/tests/test_timing_report.py
+++ b/positronic/cli/eval/tests/test_timing_report.py
@@ -1,5 +1,6 @@
 import json
 
+import pos3
 import pytest
 
 from positronic.cli.eval.timing_report import (
@@ -9,6 +10,7 @@ from positronic.cli.eval.timing_report import (
     _read_spans_dir,
     _read_stats_dir,
     _render,
+    timing_report,
 )
 from positronic.simulator.env_server.telemetry import ENV_PROCESS
 from positronic.telemetry import (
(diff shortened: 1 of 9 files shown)


vertix:
@codex review

PR #618, opened 2026-08-12
Keep an aborted episode on disk, unfinished

v-positronic:
## What

`DiskEpisodeWriter.abort()` no longer deletes the episode directory. It finalizes the signal writers and leaves what was recorded where it is, unfinished.

## Why

The episode a run was making when it stopped — crashed, ended with the episode still open, abandoned by the operator — was the one recording nobody could look at: `abort()` was `shutil.rmtree`, and `DsWriterAgent`'s teardown calls it on whatever is open. That is the case where the bytes are worth most, and they were the only ones deleted.

## How it stays out of the dataset

Nothing new. `_aborted` already makes `__exit__` return before the metadata block, so the episode gets no `meta.json` and keeps the `.unfinished` marker the writer put there at the start — which is exactly what the readers key on today: `LocalDataset` skips a directory holding it, `DiskEpisode` refuses to open one. Out of the dataset, still on disk.

The one adjacent change: the signal writers are finalized (`__exit__`) rather than aborted, because a signal writer's own `abort()` unlinks what it wrote — the thing being kept.

## What this costs

The directories accumulate and nothing sweeps them, so a run's output directory is not tidy after a crash. They also upload with the rest of the run, which is accepted: every reader that opens the dataset there ignores them, so what reaches a prefix is an extra directory nobody opens.

## Verification

`uv run pytest positronic/dataset positronic/policy positronic/server positronic/tests` — 510 passed, 4 skipped.

Two tests changed to the new contract (an aborted writer keeps its directory, marker and no `meta.json`; a context that raises leaves the same), and one added: an aborted episode is absent from `LocalDataset` while its directory is present.

### DIFF
--- a/positronic/dataset/README.md
+++ b/positronic/dataset/README.md
@@ -165,7 +165,7 @@ class EpisodeWriter:
     def __exit__(self, exc_type, exc, tb) -> None:
         ...
 
-    # Abort the episode; underlying writers are aborted and the `Episode` directory is removed
+    # Abort the episode: it stays on disk, unfinished, and never joins the dataset
     def abort(self) -> None:
         pass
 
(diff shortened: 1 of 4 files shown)


vertix:
@codex review

PR #621, opened 2026-08-13
Pass auth headers to every endpoint a sampled run reaches

v-positronic:
`production` takes `headers` and hands them to every `RemotePolicy` it builds, so a set of endpoints behind one authenticating proxy can be sampled in a single run. Two tests pin it: the headers reach every endpoint, and the default still reaches none.

Until now the multi-endpoint preset was the one path that dropped them. `RemotePolicy` has always taken `headers` as a keyword-only argument, and `authed_remote` / `nebius_remote` pass it, but `production` built `RemotePolicy(url, recording_dir=recording_dir)` — so a run against endpoints that need a credential had no way to present one. It failed in whichever direction the caller chose: unauthenticated, a 401 per episode and nothing recorded; or, when the caller passed the credential, it did not configure at all —

```
configuronic.config.ConfigError: Error instantiating "policy":
production() got an unexpected keyword argument 'headers'
```

That is the shape every rig round of the wooden-spoon task uses: several checkpoints behind one balanced sampler, one key/secret pair in front of all of them, the operator scoring episodes without knowing which endpoint she is watching. `phail_multiple` is `production.override(...)`, so it inherited the gap, and nothing else covered the case — the single-endpoint path passes no headers either, and `authed_remote` carries a bearer token rather than a header pair.

One set of headers goes to all endpoints because that is what one proxy in front of several endpoints means: the credential is the same for each. Endpoints wanting different credentials stay separate policies composed through `sample`, where per-policy configuration already belongs. The docstring says so, so the boundary does not have to be inferred. Existing callers are untouched — the argument defaults to none.

The second test is the one worth keeping honest about: it fails if the change ever starts sending headers a caller did not ask for, which is the opposite error and the easier one to ship.

Verification: both tests pass. The same patch applied to a rig's checkout took a two-endpoint blind batch from refusing at config time to launching and authenticating — the proxy accepts the pair and no longer 401s.

Refs: Positronic-Robotics/internal#411

### DIFF
--- a/positronic/cfg/policy.py
+++ b/positronic/cfg/policy.py
@@ -52,10 +52,11 @@ def balanced(balance: int):
     return BalancedSampler(balance=balance)
 
 
-@cfn.config(endpoints={}, weights={}, recording_dir=None, sampler=None, group_fields=None)
+@cfn.config(endpoints={}, weights={}, headers=None, recording_dir=None, sampler=None, group_fields=None)
 def production(
     endpoints: dict[str, str],
     weights: dict[str, float],
+    headers: dict[str, str] | None,
     recording_dir: str | None,
     sampler: Sampler | None,
     group_fields: list[str] | None,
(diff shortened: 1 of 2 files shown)


vertix:
@codex review

v-positronic:
Closing unmerged: after the sampler migration this change has nothing left to do.

`production` moves out of this library into platform's `rollouts_console` package (Positronic-Robotics/internal#440 D1/D2), and the party that builds the `RemotePolicy` is the party that forwards its headers. `RemotePolicy` has always accepted `headers` keyword-only — `authed_remote` and `nebius_remote` pass it — so once the preset is platform's, forwarding them needs no change here at all. Merging this would land library code the migration deletes.

The gap it was opened for is real and is being closed on the rig immediately (Positronic-Robotics/internal#421) and durably by that migration's PR2, which carries both these tests: the headers reach every endpoint, and the default reaches none.

Nothing in the reasoning of this change was wrong — only its home.

PR #630, opened 2026-08-15
Say whose job it is to keep the submitted image working

v-positronic:
The platform pulls a submitted policy image **as given, when the run starts**, and evaluates what it finds. It does not rewrite the reference and does not keep a copy of the image.

Two consequences a submitter cannot read off this package today, and both are theirs to avoid:

- an image deleted, made private, retagged or rate-limited between submission and execution fails that run;
- an image whose contents moved under a tag is scored as it stands at pull time.

Said in the two places the fact is reachable without knowing to look for it: on `PolicyImage`, which is the type they hold, and in `client/README.md` beside the `eval run` example they copy. Pinning a digest is the answer to both, and is now stated as such rather than implied.

The client version goes 0.2.0 → 0.2.1 with the root's pin, per the repo's version-bump gate.

Nothing here changes behaviour. The platform side of the same contract is `Positronic-Robotics/platform#18`, which stopped rewriting the reference and now records the resolved digest as evidence rather than imposing it; that PR pins the client by git rev, so it is unaffected by this version bump until someone moves the rev.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -36,6 +36,11 @@ uv run positronic eval list
 uv run positronic eval cancel --submission-id=<hex id>
 ```
 
+The image is pulled as given, when the run starts, and evaluated as found. Keeping it pullable, and
+holding what you want evaluated, is yours until the run finishes: one that has gone away by then
+fails the run, and one whose tag moved is scored as it stands at pull time. Pin a digest for a fixed
+image.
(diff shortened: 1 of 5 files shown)


vertix:
@codex review

vertix:
@codex review

PR #643, opened 2026-08-18
Add the fault guard to a declared stack that carries none

v-positronic:
An arm that faults reports its status and no `q`/`dq`/`ee_pose`. A rig-side stack that reads
those by name raises on the control loop's thread, which ends the run.

Both of today's lab rollouts died that way, the second at 8 of 30 episodes:
`KeyError: 'robot_state.ee_pose'` in `TemporalStack._Session.__call__`.

All three Runway endpoints declare, verbatim in every episode's `static.json`:
`temporal_stack | chunked_schedule | restrict_image_size` — no `stop_on_fault` on any of
them, so nothing answered that observation.

## What changes

`RemotePolicy._resolve_stack` prepends `StopOnFault()` when the declared stack contains
none, so the guard sits outside everything the handshake declared and answers the
pose-less observation before any declared wrapper sees it. A declared `stop_on_fault` is
left exactly as declared — same wrappers, same order.

This deliberately bends the documented contract that "the rig runs what the handshake
declares and nothing else", so every injection logs at WARNING naming the endpoint. A rig
silently running a stack the partner did not declare is the thing that must not become
invisible.

## Tests

- `test_a_declared_stack_without_the_fault_guard_is_given_one` — fails without the change
  with the production `KeyError` at `wrappers.py:196`, passes with it.
- `test_a_declared_fault_guard_is_built_exactly_as_declared` — pins the boundary: a stack
  declaring the guard is built as declared, so nothing is added and the declaration keeps
  deciding where the guard sits.

`positronic/tests/test_keys.py` + `positronic/offboard/tests/` + `positronic/policy/tests/`:
305 passed, 1 skipped. CI green on `175ff1a1`, all 17 jobs.

## Rules check

13 rules, 2 findings, both now resolved.

`diff-comments` — applied: a comment justified a field by naming a distant consumer.

`hardcoded-keys` on the test's `robot_state.ee_pose` — I first declined this, reasoning that a
fixture should spell the wire name rather than import a constant, since a shared constant makes
client and wire agree whatever the value. This repository settles that question the other way
and enforces it mechanically: `positronic/tests/test_keys.py::test_no_guarded_key_literals`
fails on exactly that literal. The repo's own executable rule wins, so the test now imports
`keys.EE_POSE`. The other accepted literals are unguarded handshake field names, which that
test does not cover.

## Deployment note

The lab rig cannot run this commit: `#641` removed `positronic.policy.harness.Directive`,
which the rollouts console's `eval_ui` imports, so a console pinned to `main` dies at
launcher import. The rig runs `14354007f6dca7b435280c06144fc11d26e8e116` (branch
`rollouts/inject-stop-on-fault-rig`) on the pre-`#641` base. `positronic/policy/remote.py` is
byte-identical between that commit and this PR's head; they differ only in the one test line
above, which the rig does not import. The console has to be updated for the `#641` API change
before its pin can follow main.

Not for merge without a named human's go.

Ticket: Positronic-Robotics/internal#558

### DIFF
--- a/positronic/offboard/tests/test_remote_policy.py
+++ b/positronic/offboard/tests/test_remote_policy.py
@@ -9,7 +9,7 @@ from websockets.exceptions import InvalidStatus
 from websockets.http11 import Response
 
 from positronic import keys, telemetry, telemetry_keys
-from positronic.drivers.roboarm import command
+from positronic.drivers.roboarm import RobotStatus, command
 from positronic.offboard.client import DEFAULT_INFER_TIMEOUT, InferenceClient, _ConnectRetries
 from positronic.policy import RemotePolicy
 from positronic.policy.codec import ActionHorizon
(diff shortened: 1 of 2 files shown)


v-positronic:
Closing — superseded by positronic#646, which fixes the root cause: `Serializers.robot_state` was dropping `.q`/`.dq`/`.ee_pose` for a faulted arm, so the pose was never actually missing and `temporal_stack` now has nothing to raise on. Injecting a guard the handshake did not declare was the right thing while a round was stuck on the rig this morning; it is not the right thing going forward, since it makes the rig run a stack the partner did not serve.

The deployed revision `14354007` (this branch's change on top of `0908680f`) is what the 18 Aug round ran from episode 9 onward, and it is pinned by platform#166. That pin predates positronic#646, so the rig does not yet carry the root fix — re-pinning the console to a `main` revision containing it is the follow-up, tracked on Positronic-Robotics/internal#558.

PR #653, opened 2026-08-21
Recover from an aborted homing move instead of leaving the arm short

v-positronic:
A `cartesian_reflex` aborts the vendor goal in flight. `await_goal` read the abort and raised, so the move that was meant to put the arm somewhere known simply did not happen — the arm stayed where the reflex stopped it.

It now clears the arm's error and commands the same target again, up to `_MOVE_ATTEMPTS` (3).

## Why the recovery belongs to the driver

The vendor is explicit that it does not do this itself. From `positronic-franka`'s `set_target_joints`:

> Clearing is the caller's call: a reflex means the arm hit something, and only the caller knows whether resuming is safe.

So an aborted goal leaves two facts standing: the arm holds an error, and a target commanded under one is refused outright. Clearing then re-commanding is the documented way back, and nothing was doing it.

## Recoverable vs terminal — where the distinction lives

Not in `GoalStatus`, which has one failure value: `NONE`, `IN_FLIGHT`, `REACHED`, `ABORTED`. Not in `Goal.reason` either — libfranka's own free text, carrying a reflex, a rejected plan, a lost connection and an expired deadline alike.

It lives in `recover_from_errors() -> bool`, documented as *"returns True if cleared"* and discarded at every call site until now. So the code asks it rather than classifying the reason:

- it clears → command the target again;
- it will not clear → raise at once, since the arm cannot be commanded either way;
- it keeps clearing and keeps aborting → the attempts run out and the raise stands.

The re-command goes through `command_target`, so the control law is re-applied with the target rather than left to whatever the abort dropped — the same pairing #651 requires of a first command.

## The bound

Both callers already pass a time-based `should_stop` (a travel deadline, or `park_timeout_s`), and a reflex trips in milliseconds — time alone would allow hundreds of retries into whatever the arm is against. The attempt count is what stops that: enough for a trip on the way past something, few enough that an arm genuinely held against something gives up rather than driving at it repeatedly with an operator nearby.

Homing still raises when the arm does not arrive; it gets a few chances to arrive first.

## Scope

`Robot.run` already logs and carries on at both homing call sites, so this does not change whether a failed home ends a run — it changes whether the home happens.

`park` used to clear the error itself before its move, because a reflex during that move ended the park. `await_goal` clears now, so that call had one owner too many and is gone; a park that starts on an errored arm spends its first attempt on the refusal and no wall clock.

## Tests

- `test_a_reflex_during_the_opening_home_is_cleared_and_the_arm_sent_again` — the incident. Against the current driver the arm finishes at `JOGGED`, logging `Homing failed … motion aborted by reflex! ["cartesian_reflex"]`; with the change it reaches `HOME`, with a clear between the two commands.
- `test_a_move_that_keeps_aborting_gives_up_rather_than_driving_at_it_again` — the other boundary: exactly `_MOVE_ATTEMPTS` commands, then the raise. Without it the retry is bounded only by the deadline.
- `test_a_move_gives_up_at_once_on_an_error_the_arm_will_not_clear` — a terminal fault still surfaces, on the first abort, having commanded once.

`FakeArm.recover_from_errors` returned `None`, which is not what the vendor returns; it now returns a `bool` and clears the fault flag, with `recovers` to script an arm that will not.

Lint, format, the basedpyright ratchet and the 251 tests under `positronic/drivers/` and `pimm/` pass.

## Not shipped here

The rig runs a pinned positronic (`661cd32a`), which predates #650 — there the raise escapes `Robot.run` and takes the whole World down, which is how two rounds died on 21 August at 6 of 30 and 17 of 24 episodes. Reaching the rig needs a later pin bump in `platform`, separately.

Ticket: Positronic-Robotics/internal#620

### DIFF
--- a/positronic/drivers/roboarm/franka.py
+++ b/positronic/drivers/roboarm/franka.py
@@ -110,8 +110,8 @@ class _Arm(DriverRun[command.CommandType]):
     _MOVE_GRACE_S = 5.0
     # Parking publishes nothing, so it comes back only as often as it needs to ask again
     _PARK_POLL_S = 0.005
-    # How many times a move the vendor aborted is commanded again. A reflex clears and the arm goes on; an arm
-    # held against something spends the attempts instead of driving at it for as long as the deadline allows.
+    # How many times a move the vendor aborted is commanded again — a bound a deadline cannot give, since a
+    # reflex trips in milliseconds
     _MOVE_ATTEMPTS = 3
 
     def __init__(
@@ -204,8 +204,7 @@ class _Arm(DriverRun[command.CommandType]):
                 stopped_short = f'the arm stopped short of its target: {goal.reason or goal.status}'
                 if attempts_left == 0:
                     raise RuntimeError(f'{stopped_short}, on all {self._MOVE_ATTEMPTS} attempts')
-                # A target commanded while the arm holds an error is refused outright, and the vendor leaves
-                # the clearing to the caller: only the caller knows whether going again is safe.
+                # A target commanded while the arm holds an error is refused outright
                 if not self.vendor.recover_from_errors():
                     raise RuntimeError(f'{stopped_short}, and its error will not clear')
                 logger.warning(f'{stopped_short}; cleared it, commanding again with {attempts_left} attempts left')
@@ -313,8 +312,6 @@ class _Arm(DriverRun[command.CommandType]):
         target = np.asarray(self._home_joints, dtype=np.float64)
         try:
             logger.info('Parking the arm at the home pose')
-            # Clears an error the run ended in, so the first attempt is a move rather than a refusal
-            self.vendor.recover_from_errors()
             deadline = self.clock.now() + self._park_timeout_s
             # The home pose is a long way off, and only the native law shapes the reference on the way there.
             outcome = yield from self.await_goal(
--- a/positronic/drivers/roboarm/franka.py
+++ b/positronic/drivers/roboarm/franka.py
@@ -110,6 +110,9 @@ class _Arm(DriverRun[command.CommandType]):
     _MOVE_GRACE_S = 5.0
     # Parking publishes nothing, so it comes back only as often as it needs to ask again
     _PARK_POLL_S = 0.005
+    # How many times a move the vendor aborted is commanded again. A reflex clears and the arm goes on; an arm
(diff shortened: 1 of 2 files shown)


v-positronic:
Closing: the fix is moving to the driver's control law rather than a retry.

A reflex during a home will be prevented rather than recovered from — homing switches to impedance mode so it does not trip in the first place. That addresses the cause; this branch addressed the consequence, and only for one of the two paths that reach it.

What this branch established, kept here so it is not re-derived:

`_Arm.await_goal` commands its goal exactly once and raises the moment `goal.status` leaves `IN_FLIGHT`. `move_to`'s loop calls `recover_from_errors()` when it sees `st.error != 0`, which clears the FAULT and does not re-issue the MOTION, so an aborted goal stays aborted and the next poll raises. `park()` already anticipates half of this — it clears errors once before the move, commented "a reflex during the move ends the park" — and nothing handles a reflex during.

The recoverable/terminal split is not in `GoalStatus` (one failure value) and not parseable from `Goal.reason` (libfranka free text); it is `recover_from_errors() -> bool`, whose return the driver discarded at every call site.

Measured occurrence, 24 August, run `blind_20260824-122151`, which lost 8 episodes of a partner round:

- episode closes at 15:45:09.319 with the arm at `q = [-0.139, -0.06, 0.279, -2.585, -0.141, 2.185, -0.021]`, `status = AVAILABLE`, `dq ~ 0`, end-effector at `z = 0.160 m`
- the home to `[0.0, -0.31, 0.0, -1.65, 0.0, 1.522, 0.0]` asks j4 for +53.6 deg and j6 for -38.0 deg — a large elbow lift from a low pose, at `relative_dynamics_factor = 0.2`
- `cartesian_reflex` at 15:45:09.659, 0.34 s in
- the wrist and exterior recordings for that episode show the gripper down inside the transparent tote among the spoons, so the lift starts through them
- the teardown park drove the same arm to the same target through the same function 6 s later and succeeded, so the condition was transient

That last point is why a retry would have worked, and the pose is why impedance is the better answer: the arm homes out of a tote with items around the fingers, and the trip is a property of that path rather than of the code.

Not covered by this branch and still open: the gripper reaches its target through `Moves`/`grip_setpoint`, not `await_goal`, so nothing here would have prevented `blind_20260824-112352` losing 4 episodes to `TimeoutError: stopped at 0.894, short of 0.0`.

Context and history: Positronic-Robotics/internal#620.

PR #667, opened 2026-08-25
Match the certificate subject however `openssl` spaces it

v-positronic:
## What

`test_certificate_carries_a_subject_a_long_host_name_would_overflow` asserted the substring `CN = positronic-server`. Assert the common name with a pattern that tolerates either spacing.

## Why

That spacing is OpenSSL's `oneline` name format. LibreSSL, which macOS ships, prints the same subject as `CN=positronic-server`, so the assertion pinned the printer rather than the certificate. `core (macos-latest, 3.13)` failed on 5da7d2db6 while every Linux job passed; main has been red since.

The two forms, both now matched:

| Runner | `openssl x509 -noout -subject` |
|---|---|
| ubuntu-latest (OpenSSL 3.0.13) | `subject=CN = positronic-server` |
| macos-latest (LibreSSL) | `subject=CN=positronic-server` |

Reading the subject through `-nameopt` would also fix it, but that flag's option set is not identical across the two implementations, and a pattern needs no support from either.

## Test

`uv run --locked pytest --no-cov` — 1643 passed, 8 skipped, on Linux/OpenSSL 3.0.13. The pattern was also checked against the macOS output captured from the failing run.

### DIFF
--- a/positronic/server/tests/test_positronic_server.py
+++ b/positronic/server/tests/test_positronic_server.py
@@ -93,7 +93,8 @@ def test_certificate_carries_a_subject_a_long_host_name_would_overflow():
     assert len(host.encode()) > 64
     text = _certificate_text([host], '-subject', '-ext', 'subjectAltName')
 
-    # OpenSSL prints the subject as `CN = value`, LibreSSL as `CN=value`.
+    # Every implementation spaces the subject its own way: `CN = value` on OpenSSL 3.0, `CN=value` on
+    # OpenSSL 3.6, `/CN=value` on LibreSSL.
     assert re.search(r'CN\s*=\s*positronic-server', text)
     assert f'DNS:{host}' in text
 
--- a/positronic/server/tests/test_positronic_server.py
+++ b/positronic/server/tests/test_positronic_server.py
@@ -93,7 +93,8 @@ def test_certificate_carries_a_subject_a_long_host_name_would_overflow():
     assert len(host.encode()) > 64
     text = _certificate_text([host], '-subject', '-ext', 'subjectAltName')
 
-    assert 'CN = positronic-server' in text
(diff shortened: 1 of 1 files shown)


vertix:
I pushed one commit to this branch (`ade4d533`), because #665 and every other branch cut from `main` inherits this red build.

It changes the inline comment only. The assertion is yours and is unchanged.

The reason: the split is not OpenSSL against LibreSSL. I measured three formats on one macOS machine.

| implementation | printed subject |
|---|---|
| OpenSSL 3.0 (the ubuntu runners) | `subject=CN = positronic-server` |
| OpenSSL 3.6.1 (homebrew, first on `PATH`) | `subject=CN=positronic-server` |
| LibreSSL 3.3.6 (`/usr/bin/openssl`) | `subject= /CN=positronic-server` |

So `macos-latest` prints `CN=` because its `openssl` is a current OpenSSL, and not because it is LibreSSL. Your pattern matches all three, including the LibreSSL form the branch had not named.

The body of this PR still states the two-implementation reading. Correct it if you want the record straight; the code now names all three.

vertix:
Closing in favour of #669, which fixes the same failure by reading the certificate instead of the printout.

Your diagnosis was right, and #669 carries it further: the tests load the PEM and read the subject and the alternative names, so no spacing can break them. The alternative-name tests had the same defect — `_certified_ips` scraped addresses out of the extension text — and they are fixed too. `cryptography` joins the `dev` group to allow it.

One correction from this branch that #669 keeps: the split is not OpenSSL against LibreSSL. OpenSSL 3.0 prints `CN = value`, OpenSSL 3.6 prints `CN=value`, and LibreSSL prints `/CN=value`. `macos-latest` fails because its `openssl` is a current OpenSSL.

Reopen this if #669 does not land.

